PR #3, opened 2026-03-26
Cap max fee at 5% to prevent collateral rug via fee manipulation

LandynDev:
Previously fee_divisor minimum was 2 (allowing up to 50% fee on miner collateral per swap). Now enforced at minimum 20 (max 5%) in both the contract and CLI.


anderdc (approved):
LGTM

PR #6, opened 2026-04-01
Verify dest transaction sender matches miner's posted address

LandynDev:
**Verify dest transaction sender matches miner's posted address**

## Summary
- **Security fix**: The dest-side swap verification only checked that the correct amount reached the user's address by tx hash — it did not verify the transaction came FROM the miner's posted address. A miner could watch for any third-party tx to the user's dest address of the correct amount and claim that tx hash as their fulfillment without actually sending funds.
- **Contract**: Added `miner_dest_address` field to `SwapData`
- **Validator**: `chain_verification.py` now validates `tx_info.sender` against `swap.miner_dest_address` on the dest leg, rejecting mismatches

## Test plan
- [x] `cargo contract build` compiles with new field
- [x] `ruff check allways/ neurons/` passes
- [x] Restart dev environment and run E2E suite 02 (swap lifecycle)
- [x] Verify validator logs show sender validation on dest-side confirmation

### DIFF
--- a/allways/classes.py
+++ b/allways/classes.py
@@ -62,6 +62,7 @@ class Swap:
     user_source_address: str
     user_dest_address: str
     miner_source_address: str = ''
+    miner_dest_address: str = ''
     rate: str = ''
     source_tx_hash: str = ''
     source_tx_block: int = 0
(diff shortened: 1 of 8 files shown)


plind-junior:
What is it for? Specifically for which issue?

LandynDev:
> What is it for? Specifically for which issue?

It's specifically for this edge case:
image

anderdc (approved):
LGTM

PR #10, opened 2026-04-06
feat: Prevent miner slashing on API outages by distinguishing tx-not-found from provider-unreachable

Crystora:
## Prevent miner slashing when chain provider is unreachable

Right now `verify_transaction()` returns `None` for both "transaction not found" and "API is down." The validator treats both the same — assume the miner never sent funds, eventually timeout and slash.

This means a miner who fulfilled correctly can lose collateral just because Blockstream was temporarily unreachable.

### What changed

Added a `ProviderUnreachableError` exception that providers raise on transient failures (connection errors, timeouts, HTTP 5xx). The existing `Optional[TransactionInfo]` return stays for found/not-found.

Three call sites catch this exception:

- **Pending confirms** — keeps the item for retry instead of dropping it
- **Fulfilled verification** — marks the swap as uncertain so it isn't timed out
- **Timeout check** — skips swaps where verification was inconclusive this cycle

BTC RPC → Blockstream fallback is unchanged. If RPC says "not found", Blockstream is still tried.

### Before

```
API down → verify returns None → timeout → miner slashed
```

### After

```
API down → ProviderUnreachableError → defer timeout → retry next cycle
```

Unrelated method extractions (`_match_transfer`, `_rpc_resolve_sender`, `_blockstream_calc_confirmations`) have been moved to a separate PR (#18) to keep this diff focused on the fix.


Crystora:
@anderdc @LandynDev Ready for review. Please give me any feedbacks.

LandynDev (changes requested):
##### 1. The fix is bigger than it needs to be

The actual behavioral change is ~15 lines in `forward.py`.  
The other ~500 lines are a new type system (`VerifyResult` + `VerifyOutcome` + factory methods + custom exceptions + retry machinery).

The consumers only care about two questions:
- "is it verified?"
- "was there an error?"

The intermediate enums and factory methods add indirection without clarity.

**Simpler approach:**  
Have `verify_transaction()` raise a `ProviderUnreachableError` on transient failures, keep the existing `Optional[TransactionInfo]` return for found/not-found, and catch the exception in the 3 call sites that care.

→ ~50 line PR instead of a large refactor.

##### 2. Unrelated refactors should be a separate PR

The extractions:

- `_match_transfer`
- `_rpc_resolve_sender`
- `_blockstream_calc_confirmations`

are fine on their own, but they inflate this diff and make the behavioral change harder to review.

Splitting them into a separate PR would be my preference so this work stays modular and focused on the fix. Easier for me to review

---

##### 3. BTC RPC → Blockstream fallback behavior changed

Current behavior:

```
RPC says "not found"
      ↓
Still try Blockstream
      ↓
May find mempool tx not yet indexed by the node
```

New behavior in the PR:

```
RPC says "not found"
      ↓
STOP
```

Fallback only happens on error, not on "not found".

**Question:** Is that intentional? This change could cause missed verifications.

Crystora:
All three review points addressed:
1. Smaller fix - Dropped the VerifyResult / VerifyOutcome / retry machinery. verify_transaction() still returns Optional[TransactionInfo] for found vs not found; transient failures raise ProviderUnreachableError.

2. Refactor split out - _match_transfer, _rpc_resolve_sender, and _blockstream_calc_confirmations are not in this PR; they’re in a separate PR (#18) so this diff stays behavioral-only.

3. BTC RPC → Blockstream fallback - Unchanged: RPC still returns None for “not found” (and errors), then Blockstream still runs. Only Blockstream treats real “not found” (404) as None vs network/HTTP failures as unreachable.

Crystora:
@LandynDev Thanks for the detailed review. I’ve addressed all. Could you please take look once more?

LandynDev (changes requested):
We should add asymmetric protection

This is additional to your idea and just protecting the other side, bc we're shielding miners from wrongful slashing during chain outages, but the user-entry path is still exposed to the same trigger with a strictly worse outcome:

User sends source funds → validator can't verify → `reserved_until` lapses → contract releases the miner → user's funds end up sitting at the miner's address with no contract claim.

Beyond the passive fund-loss case, the asymmetry also feels like real **attack surface** during any chain halt or degradation — miners could post bait rates knowing verification is blind, and users could enter swaps specifically because lookups can't happen.

Both sides should get deferral protection

**What we should do:**

Mirror the existing `extend_reservation` block in `_process_pending_confirms` onto the `ProviderUnreachableError` path — same `request_hash` + `vote_extend_reservation` call, no new primitives needed.

We might want to pull the verified-path extend block into a small helper and share it across both sites

----

I think that'd be great to add here

Crystora:
@LandynDev I updated correctly.

LandynDev (approved):

Crystora:
Thanks!

PR #13, opened 2026-04-07
Add swap quote and swap resume CLI commands

### LINKED ISSUE #12: Add swap quote and swap resume CLI commands
## Summary

Two new CLI commands under `alw swap` to improve user experience:

### `alw swap quote`
Preview rates and estimated receive amounts **before committing** to a swap.

```
alw swap quote --from btc --to tao --amount 0.1
```
```
 #  UID  Rate (TAO/BTC)  You Receive          Collateral
 1    4  345             34.15500000 TAO      2.0000 TAO
 2   12  340             33.66000000 TAO      1.5000 TAO
                                    (after 1% fee)
```

**Why:** Currently there's no way to preview what you'd receive for a specific amount without entering the full `alw swap now` flow. `alw view rates` shows raw rates but doesn't calculate amounts or apply fees.

### `alw swap resume`
Recover an interrupted swap — same wizard UX as `alw swap now`, same agent-friendly flag support.

```
# Interactive (wizard mode)
alw swap resume

# Non-interactive (agent mode)  
alw swap resume --source-tx-hash abc123 --yes
```

**Why:** If a user's terminal disconnects mid-swap or a BTC send fails, they're currently told "Resume with: `alw swap post-tx <tx_hash>`" — but that requires remembering the tx hash and context. `alw swap resume` reads the existing `~/.allways/pending_swap.json` and handles all recovery scenarios:
- Reservation expired → auto re-reserves same miner
- Miner offline → finds alternative at same-or-better rate
- Source tx already sent → skips to confirm flow
- Swap already on-chain → skips to watch mode

## Implementation Details

- **No new infrastructure** — both commands use only existing data and contract queries
- `quote`: ~130 lines — reads `read_miner_commitments()`, calculates via `calculate_dest_amount()` + `apply_fee_deduction()`
- `resume`: ~330 lines — orchestrates existing `broadcast_reserve_with_retry()`, `sign_and_broadcast_confirm()`, `poll_for_swap_creation()`
- Both follow existing CLI patterns (rich-click, `get_cli_context()`, global flags)

eureka0928:
Closes #12

## Summary

Two new CLI commands under `alw swap` to improve user experience:

- **`alw swap quote`** — Preview rates and estimated receive amounts before committing to a swap. Shows all available miners with calculated receive amounts after fees for a given input amount.
- **`alw swap resume`** — Recover an interrupted swap flow. Reads saved pending state and handles all failure scenarios: expired reservations, offline miners, already-on-chain swaps.

## What Changed

- `allways/cli/swap_commands/quote.py` — New command (~130 lines)
- `allways/cli/swap_commands/resume.py` — New command (~330 lines)
- `allways/cli/swap_commands/__init__.py` — Register both commands under `alw swap`
- `.gitignore` — Add `.venv/`

## Design

Both commands follow the existing CLI patterns and reuse existing infrastructure:

**Quote** uses `read_miner_commitments()` + `calculate_dest_amount()` + `apply_fee_deduction()` — purely read-only, no wallet needed.

**Resume** follows the same two philosophies as `alw swap now`:
1. **Wizard mode** — walks user through recovery step by step, auto re-reserves if expired, finds alternative miners if original went offline
2. **Agent mode** — fully scriptable via `--source-tx-hash` and `--yes` flags

### Resume handles these recovery paths:
| Scenario | Behavior |
|---|---|
| Swap already on-chain | Skip to watch mode |
| Reservation still active | Proceed to tx submission |
| Reservation expired, miner available | Auto re-reserve same miner |
| Reservation expired, miner unavailable | Find alternative at same-or-better rate |
| No pending swap | Clear message, suggest `alw swap now` |

## Usage

```bash
# Quote
alw swap quote --from btc --to tao --amount 0.1
alw swap quote --from tao --to btc --amount 50

# Resume (interactive)
alw swap resume

# Resume (non-interactive / agent)
alw swap resume --source-tx-hash abc123... --yes
```

## Test Plan

- [ ] `alw swap quote --from btc --to tao --amount 0.1` shows rates table with receive amounts
- [ ] `alw swap quote` with invalid chain shows error
- [ ] `alw swap resume` with no pending swap shows clear message
- [ ] `alw swap resume` with active reservation proceeds to tx prompt
- [ ] `alw swap resume --source-tx-hash <hash> --yes` works non-interactively
- [ ] Existing tests pass (`pytest tests/test_rate.py tests/test_chains.py`)
- [ ] `ruff check` and `ruff format` pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 venv/
+.venv/
 allways-venv/
 .env
 *.env
--- a/allways/cli/swap_commands/__init__.py
+++ b/allways/cli/swap_commands/__init__.py
@@ -4,6 +4,8 @@ from allways.cli.swap_commands.collateral import collateral_group
 from allways.cli.swap_commands.miner_commands import miner_group
 from allways.cli.swap_commands.pair import post_pair
 from allways.cli.swap_commands.post_tx import post_tx_command
+from allways.cli.swap_commands.quote import quote_command
+from allways.cli.swap_commands.resume import resume_command
 from allways.cli.swap_commands.status import status_command
 from allways.cli.swap_commands.swap import swap_group
 from allways.cli.swap_commands.view import view_group
@@ -11,8 +13,10 @@ from allways.cli.swap_commands.view import view_group
 # Register post under the miner group
 miner_group.add_command(post_pair, 'post')
 
-# Register post-tx under the swap group
+# Register post-tx, quote, resume under the swap group
 swap_group.add_command(post_tx_command, 'post-tx')
+swap_group.add_command(quote_command, 'quote')
+swap_group.add_command(resume_command, 'resume')
 
 
 def register_commands(cli):
--- a/allways/cli/swap_commands/helpers.py
+++ b/allways/cli/swap_commands/helpers.py
@@ -9,7 +9,7 @@ from typing import Optional, Tuple
 import bittensor as bt
 from rich.console import Console
 
-from allways.classes import SwapStatus
+from allways.classes import MinerPair, SwapStatus
 from allways.commitments import parse_commitment_data, read_miner_commitment, read_miner_commitments  # noqa: F401
 from allways.constants import CONTRACT_ADDRESS as DEFAULT_CONTRACT_ADDRESS
 from allways.constants import NETUID_FINNEY, TAO_TO_RAO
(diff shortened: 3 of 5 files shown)


fury0928:
Hi @LandynDev would you review this PR?
Thank you for your time

LandynDev (changes requested):
A few things we'd want to tighten up before merging
---

### Resume: on-chain swap scan performance

`get_miner_active_swaps()` calls `get_active_swaps()` which iterates through all swap IDs.  On mainnet with lots of historical swaps this could be really slow for what should be a quick recovery command.

We might wanna add a more targeted lookup — we already know the `miner_hotkey` from pending state, so we could:

- Cap the scan window, **or**
- Check `get_miner_has_active_swap` first and only do the full scan if that returns true (this is probably better bc if the miner doesn't have an active/pending swap then there's nothing to resume)

---

### Resume: expired reservation = nothing to resume

Like the above comment, if the reservation expired and the user already sent funds, there's not really a resumable state — the miner could have been taken by someone else, rate could have changed, etc.

We should make it clear in this case that there's nothing to resume and guide them to start fresh with:

`alw swap now`

Right now it silently tries to re-reserve which could lead to confusing outcomes.
If the reservation expired, there's nothing to resume - they should do `alw swap now`

---

### Resume: validator discovery ordering

Right now validators are discovered *after* the user enters their tx hash.
If no validators are found, they've already committed info but hit a dead end.
In `swap now`, validator discovery happens during reservation so failures surface **before** the user takes any irreversible action.

We'd want to match that pattern here — discover validators **before** prompting for the tx hash.

---

### Resume: empty `dest_address` on re-reservation

When building the `MinerPair` for re-reservation (line 178), `dest_address` is set to `''`.

It works today since `broadcast_reserve_with_retry` doesn't use it, but it's a footgun if anything downstream ever reads it.

We might wanna just pass the actual `dest_address` through.

---

Overall solid contribution — the `swap now` command is the reference point for how these should feel, both for interactive users and agents.

Once these are addressed this should be good to go.

eureka0928 (commented):
All four items addressed:

**1. On-chain scan performance** — Now gates with `get_miner_has_active_swap()` (cheap bool) before doing the full `get_miner_active_swaps()` scan.

**2. Expired reservation = nothing to resume** — Removed the re-reserve logic entirely. If the reservation expired, we clear the pending state and direct them to `alw swap now`. Much cleaner — no confusing silent re-reservation.

**3. Validator discovery ordering** — Moved `discover_validators()` before the tx hash prompt. If no validators are found, we fail before the user enters anything irreversible.

**4. Empty dest_address** — Removed the `MinerPair` reconstruction entirely since re-reservation is no longer part of resume. The `dest_address=''` footgun is gone.

The command is now much simpler: show summary → check if already on-chain → verify reservation is active → discover validators → prompt for tx hash → confirm. ~200 lines down from ~330.

fury0928:
@LandynDev would you take another look on this? I addressed all
Thank you

LandynDev (approved):

PR #15, opened 2026-04-07
Rate-limit user-facing axon endpoints to block request floods before they hit chain queries

Crystora:
## Rate-limit swap reserve and confirm endpoints

Right now `blacklist_swap_reserve` and `blacklist_swap_confirm` are pass-throughs - they always return `(False, 'Passed')`. Every request goes straight to the forward handler, grabs `axon_lock`, and fires off chain RPC calls. Nothing stops a single IP from flooding these endpoints and blocking all other traffic.

This adds a simple per-IP rate limiter (10 req/min sliding window) that rejects floods in the blacklist phase, before any expensive work happens.

### Changes

- New `AxonRateLimiter` class in `allways/validator/rate_limiter.py` - thread-safe sliding window, auto-cleans stale entries
- Updated both blacklist handlers to check the rate limiter instead of passing everything through
- Added rate limit constants to `constants.py`
- Initialized the limiter on the validator in `neurons/validator.py`

### Notes

- Fails open - if IP is missing or limiter isn't set up, requests pass through as before
- MinerActivate already has hotkey auth in its blacklist, so it doesn't need this
- The dendrite IP comes from a client header so it can be spoofed, but it still raises the bar compared to zero protection

### DIFF
--- a/allways/constants.py
+++ b/allways/constants.py
@@ -41,6 +41,11 @@ SCORING_SUCCESS_EXPONENT = 8  # Harsh failure penalty: 92% → 0.51x, 96% → 0.
 RECYCLE_UID = 53  # Subnet owner UID — emissions recycled on-chain
 DAILY_EMISSION_ALPHA = 7200 * 0.41  # 2952 alpha/day (7200 blocks/day * 0.41 miner share)
 
+# ─── Axon Rate Limiting ──────────────────────────────────
+RATE_LIMIT_MAX_REQUESTS = 10  # Max requests per IP within the sliding window
+RATE_LIMIT_WINDOW_SECONDS = 60  # Sliding window duration in seconds
+RATE_LIMIT_CLEANUP_SECONDS = 300  # Purge stale IP entries every 5 min
+
 # ─── Reservation ─────────────────────────────────────────
 RESERVATION_COOLDOWN_BLOCKS = 150  # ~30 min base cooldown on failed reservation (validator-enforced)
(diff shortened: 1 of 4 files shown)


LandynDev:
Appreciate the effort, but this isn't worth fighting at the Python level imo                                                                             
                                                                                                                                                  
- **Abuse is already covered.** Contract-enforced cooldown per source address (`RESERVATION_COOLDOWN_BLOCKS * 2^(strike-1)`) plus signature proofs on both
`SwapReserve` and `SwapConfirm` already make it impossible to push bogus swaps through. Nothing in this PR changes the protocol's safety posture.
- Per-IP limiting is weak by construction. Rotating residential proxies easily bypass
- Wrong layer, real flood protection belongs at the ingress (nginx, cloud LB, fail2ban, Cloudflare), not in validator Python. Ingress-level mitigations
are harder to bypass, don't eat validator memory per caller, and compose with CIDR blocks and more specializastions

Crystora:
Thanks for the feedback. I think the cooldowns and proofs don't cover this case though:

Both swap blacklist functions are pass-throughs, so every request reaches the forward handler. The forward handler grabs axon_lock before any cooldown or proof check runs. That's the same lock handle_miner_activate uses. So flooding SwapReserve/SwapConfirm blocks miner activations - the cooldowns can't prevent that because they run after the lock is already held.
This PR rejects floods in the blacklist phase, before the lock is touched.

Would you please reopen and review this PR again?

PR #17, opened 2026-04-07
Persist scoring window and voted swaps across validator restarts

### LINKED ISSUE #16: Validator restart loses scoring window and can wipe miner weights
After a validator restart, we lose the in-memory scoring window (`SwapTracker.window`) and voted swap ids.
That means the next scoring pass runs with almost no recent history. Since scoring alpha is 1.0, this can heavily skew or wipe miner weights right after restart. This isn’t a chain-data problem — it’s local state not being persisted.

Expected behavior: restart should continue from recent scoring state, not reset it.

bitloi:
Closes #16

## Summary

- Add `ScoringWindowStore` for atomic JSON persistence of the scoring window and voted set, following the existing tmp-file-then-rename pattern used by `SwapFulfiller._save_sent_cache.`
- Wire persistence into `SwapTracker` load on `initialize()`, persist on swap resolution, window pruning, and `mark_voted().`
- Remove duplicate `_resolved_block` helper by sharing a single `resolved_block()` from the store module.
- Create store in validator init with cache path at `<neuron.full_path>/scoring_window.json`.

## Design

Persistence hooks at every state mutation point only. The store is optional (`None` disables persistence) so existing tests and standalone usage remain unaffected. On cold start, stale voted IDs are pruned against the active swap set to prevent unbounded growth.

## Test plan

- [x] `ruff check allways/validator/scoring_store.py allways/validator/swap_tracker.py neurons/validator.py` passes
- [x] `python3 -m pytest tests/test_rate.py tests/test_chains.py -q --tb=short` passes
- [x] `python3 -m pytest tests/test_scoring_store.py -q --tb=short` passes (local)
- [x] Roundtrip serialization preserves window and voted IDs
- [x] Corrupt/missing cache file falls back to empty state
- [x] Stale voted IDs pruned on restart when no active swaps exist
- [x] Window entries older than `SCORING_WINDOW_BLOCKS` pruned on load

### DIFF
--- a/allways/validator/forward.py
+++ b/allways/validator/forward.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+from dataclasses import replace
 from statistics import mean
 from typing import TYPE_CHECKING, Dict, Set, Tuple
 
@@ -228,7 +229,30 @@ async def _verify_fulfilled(
         if result:
             if voter.confirm_swap(swap.id):
                 tracker.mark_voted(swap.id)
+                _resolve_after_vote(tracker, swap, SwapStatus.COMPLETED, current_block)
                 bt.logging.success(f'Swap {swap.id}: verified complete, confirmed')
+
+
+def _resolve_after_vote(tracker: SwapTracker, swap, terminal_status: SwapStatus, current_block: int) -> None:
+    """Persist terminal outcomes promptly after a successful vote.
+
+    If the contract still returns the swap, prefer that payload. If the swap is
+    no longer queryable (resolved entries are pruned on-chain), persist a local
+    terminal snapshot to preserve scoring continuity across restarts.
+    """
+    latest = None
+    try:
+        latest = tracker.client.get_swap(swap.id)
+    except Exception as e:
+        bt.logging.debug(f'Swap {swap.id}: post-vote refresh failed, deferring resolve persistence: {e}')
+        return
+
+    if latest is not None:
+        if latest.status in (SwapStatus.COMPLETED, SwapStatus.TIMED_OUT):
+            tracker.resolve(latest)
+        return
+
+    tracker.resolve(replace(swap), status=terminal_status, current_block=current_block)
 
 
 def _extend_near_timeout_fulfilled(self: Validator) -> None:
@@ -296,6 +320,7 @@ def _timeout_expired(self: Validator, tracker: SwapTracker, voter: SwapVoter) ->
 
         if voter.timeout_swap(swap.id):
             tracker.mark_voted(swap.id)
+            _resolve_after_vote(tracker, swap, SwapStatus.TIMED_OUT, self.block)
             bt.logging.warning(f'Swap {swap.id}: timed out')
 
 
--- a/allways/validator/scoring_store.py
+++ b/allways/validator/scoring_store.py
@@ -131,8 +131,7 @@ class ScoringWindowStore:
             if isinstance(v, int):
                 voted_ids.add(v)
 
-        valid_voted_count = sum(1 for v in raw_voted if isinstance(v, int))
-        if len(window) != len(raw_window) or len(voted_ids) != valid_voted_count:
+        if len(window) != len(raw_window) or len(voted_ids) != len(raw_voted):
             self.save(window, voted_ids)
 
         if window:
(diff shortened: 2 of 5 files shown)


bitloi:
@LandynDev @anderdc Ready for review!

LandynDev:
Thanks, but this was solved and merged just before this was opened. If there's any **necessary** tweaks that should be made to the implementation, feel free to open another PR. 

Solution that was merged for this: https://github.com/entrius/allways/pull/11

bitloi:
Thanks for the clarification. I re-checked the mapping and I think there’s a mismatch:
 
- `#11` (merged) persists `PendingConfirmQueue` in SQLite (`pending_confirms.py` + `axon_handlers.py` changes).
- Issue `#16` is specifically about restart loss of `SwapTracker.window` and voted swap IDs, i.e. scoring-state persistence.
 
So while `#11` addresses validator restart durability for pending confirms, it doesn’t appear to implement the scoring-window persistence described in `#16`.
 
Would you please reopen and review this PR again?

LandynDev (changes requested):
1. Can we add tests? Can you also post results of unit tests passing after?

For example:
- round-trip save/load preserves all swap fields  
- corrupted cache file → load returns empty, no crash  
- compaction re-save when stale entries dropped (regression for the af191a0 bug)
Plus a few more covering window-age pruning and the `initialize` voted_ids intersection.

2. Can we move `resolved_block` back?

3. Can you update stale class docstring for upkeeping?

`swap_tracker.py:14–23` still says:
> *"The scoring window populates naturally as swaps complete."*
Now outdated, we can update to mention store-backed restoration

---

Other:
- Cache path uses `config.neuron.full_path` vs `PendingConfirmQueue`'s `~/.allways/validator/`. 

Good to go after

bitloi:
Implemented ✅

- Added tests for:
  - round-trip save/load preserves all swap fields
  - corrupted cache file returns empty state without crashing
  - compaction re-save when stale entries are dropped (regression for af191a0)
  - window-age pruning behavior
  - initialize-time voted_ids intersection with active swaps

- Moved the resolved-block helper back into the tracker module (`_resolved_block`).
- Updated the stale SwapTracker class docstring to mention store-backed restoration on cold start.
- Kept cache path as `<neuron.full_path>/scoring_window.json` (neuron-local runtime state).

Validation results:

- `python3 -m pytest tests/test_scoring_store.py -q --tb=short`
  - `17 passed`
- `python3 -m pytest tests/test_rate.py tests/test_chains.py -q --tb=short`
  - `37 passed`
- `ruff check allways/validator/scoring_store.py allways/validator/swap_tracker.py tests/test_scoring_store.py`
  - `All checks passed`

PunchTheDev (changes requested):
hey, I think there's still a gap in the persistence wiring that defeats the stated invariant:

- **`resolve()` doesn't persist** — `SwapTracker.resolve()` at `swap_tracker.py:70` is the primary "move a swap into the scoring window" path, called from `forward.py:250` and `:322` on every confirm and timeout vote. It mutates `window` and `voted_ids` but never calls `_persist()`. A crash between `resolve()` and the next `prune_window`/`_poll_inner` that happens to persist for another reason will lose the resolved swap — which is the exact scenario this PR exists to prevent. The docstring currently says "persistence hooks at every state mutation point," so this should be one, right?
- **`resolved_block` is still duplicated** — my last review asked to consolidate this. `_resolved_block` is back in `swap_tracker.py:194` (good) but `resolved_block` also still lives in `scoring_store.py:148` doing the same thing. Please pick one home and have the other import it.
- **missing the load-bearing regression test** — the store tests are solid, but none of them cover `resolve()` → new tracker → window restored. That's the single most valuable test for this PR's invariant and would have caught the bullet above.

LandynDev (changes requested):
Can you address requested changes left by @PunchTheDev ?
and let me know thoughts on point 1

bitloi:
> Can you address requested changes left by @PunchTheDev ? and let me know thoughts on point 1

Valid catch. There was a real crash window between voting and the next incidental persist. I fixed it.

LandynDev:
Thanks for this, at the time of your contribution we were mid-way through a planned refactoring pass, and this was naturally covered in that work by #46, which introduced `allways/validator/state_store.py`. Great contribution with clearly same vision.

PR #20, opened 2026-04-08
Harden miner fulfillment safety and recovery logging

### LINKED ISSUE #19: Miner fulfillment safety: timeout margin, terminal error classification, recovery logging
A miner can lose funds in several ways during swap fulfillment:

1. **Near-timeout sends** — Sending destination funds with only 1–2 blocks remaining before swap timeout. The `mark_fulfilled` transaction may not finalize in time, causing the swap to time out and the miner to lose both the sent funds and collateral.
2. **Futile retries** — Retrying `mark_fulfilled` that will never succeed.
3. **No recovery visibility** — When funds are sent but fulfillment fails terminally or the swap times out, there is no persistent record for the operator to act on.

bitloi:
Closes #19

## Summary

- Add timeout margin gate (`FULFILLMENT_TIMEOUT_MARGIN_BLOCKS = 3`) to prevent new sends when too few blocks remain before swap timeout. Cached sends bypass the margin to allow `mark_fulfilled` retries.
- Classify terminal `mark_fulfilled` contract errors (`SwapNotFound`, `InvalidStatus`, `NotAssignedMiner`, `MinerNotActive`) to stop futile retries.
- Write append-only JSONL recovery log for stuck-funds scenarios (terminal failures and timed-out swaps with sent funds).

## Design

The timeout margin gates new fund sends only. Once funds are already sent (cached), all downstream retry paths remain open regardless of proximity to timeout so the miner can still finalize on-chain.

## Test plan

- [x] Timeout margin blocks new sends within 3 blocks of deadline
- [x] Cached sends bypass margin and retry `mark_fulfilled`
- [x] Terminal contract errors halt retries and write recovery entry
- [x] Transient errors (`RPC_FAILURE`, `INSUFFICIENT_BALANCE`) allow continued retry
- [x] Stale cache cleanup detects timed-out swaps and logs recovery entry
- [x] Completed swaps cleaned up without false-positive recovery alerts
- [x] Focused simulation tests pass for all edge cases

### DIFF
--- a/allways/constants.py
+++ b/allways/constants.py
@@ -55,6 +55,7 @@ SWAP_FEE_PERCENT = 0.01  # display only — derived from DEFAULT_FEE_DIVISOR
 MAX_FEE_PERCENT = 0.05  # contract enforces divisor >= 20 (max 5% fee)
 MIN_COLLATERAL_TAO = 0.1  # Must be > max swap amount
 DEFAULT_FULFILLMENT_TIMEOUT_BLOCKS = 30  # ~5 min — `alw admin set-timeout`
+FULFILLMENT_TIMEOUT_MARGIN_BLOCKS = 3  # ~36s at 12s/block — avoid near-timeout sends that may fail to finalize on-chain
 DEFAULT_MIN_SWAP_AMOUNT_RAO = 100_000_000  # 0.1 TAO — `alw admin set-min-swap`
 DEFAULT_MAX_SWAP_AMOUNT_RAO = 500_000_000  # 0.5 TAO   — `alw admin set-max-swap`
 RESERVATION_TTL_BLOCKS = 30  # ~5 min — `alw admin set-reservation-ttl`
--- a/allways/miner/fulfillment.py
+++ b/allways/miner/fulfillment.py
@@ -7,8 +7,9 @@ from typing import Dict, Optional, Set, Tuple
 import bittensor as bt
 
 from allways.chain_providers.base import ChainProvider
-from allways.classes import Swap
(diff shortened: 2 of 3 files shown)


LandynDev:
hey, double-send is already prevented at `fulfillment.py:206` via the `_sent` short-circuit, with `_save_sent_cache()` persisting atomically and `_load_sent_cache()` restoring on startup. Miner stability is present and sufficient for MVP. Further and further engineering requires further and further maintenance, must choose our work carefully too

on the three concerns in #19:

- **near-timeout sends** — today it's just the bare `current_block >= swap.timeout_block` check at `fulfillment.py:85`. a cushion is a one-line add at the miners choice (could be added with a default, that's fair)
- **futile retries** — retries already happen implicitly and safely. `process_swap` returning False skips `mark_processed`, so the swap re-enters `new_pending` next poll tick; the `_sent` cache at `fulfillment.py:206` short-circuits the re-send and we retry `mark_fulfilled` with the same payload until it lands or the swap exits `Active`. that covers transient RPC, `CALL_FAILED`, and balance blips.
- **no recovery visibility** — `bt.logging.error` already hits disk and `_sent` persists atomically across restart

bitloi:
On retries — the implicit loop handles transient errors fine, but `SwapNotFound`/`InvalidStatus`/`NotAssignedMiner `from `mark_fulfilled` will never land. The miner retries that same doomed call every cycle with no escalation. That's the specific gap: funds left the wallet, contract permanently rejects the update, and the only trace is buried in rotating logs alongside every other error.

PR #23, opened 2026-04-08
Persist scoring window and voted swaps across validator restarts (#16)

itdeveloper1988:
Add ScoringWindowStore for atomic JSON persistence of the scoring window and voted set so that validator restarts no longer zero miner weights. 
Uses os.replace for cross-platform atomic rename (fixes Windows bug in fulfillment.py too). Store is optional — existing tests and standalone usage remain unaffected.


itdeveloper1988:
@LandynDev @anderdc  Could you please review my PR and give me feedback? Thanks.

PR #25, opened 2026-04-08
feat: add timeout margin gate, terminal error handling, and recovery log to miner fulfillment

itdeveloper1988:
Gate new destination sends behind a configurable block margin before swap timeout to prevent near-deadline fund loss. 
Cached sends bypass the margin so mark_fulfilled retries remain open. 
Classify SwapNotFound, InvalidStatus, NotAssignedMiner, and MinerNotActive as terminal to halt futile retries. 
Write stuck-funds events to an append-only JSONL recovery log for operator action. 
Detect timed-out swaps during stale cache cleanup and surface them in the recovery log. 
Use os.replace for cross-platform atomic file writes.


itdeveloper1988:
@LandynDev @anderdc 
Could you please review my PR and give me feedback? Thanks.

PR #27, opened 2026-04-09
Fix permanent fund lock in pending slashes

dripsmvcp:
## Problem

During a security audit of the ink! smart contract, I found a critical fund-lock vulnerability in the pending slash mechanism.

**Flow that locks funds permanently:**

1. `timeout_swap` slashes miner collateral and calls `transfer(swap.user, actual_slash)`
2. If the transfer fails (e.g. user account reaped below existential deposit), funds go to `pending_slashes` mapping and swap data is deleted (line 839)
3. User calls `claim_slash` — but if their account is destroyed/reaped, the transfer fails again. The function re-inserts the pending slash (line 916-918)
4. **No recovery path exists** — funds are permanently locked in the contract

On Substrate chains, accounts are reaped when balance drops below existential deposit. Once reaped, all subsequent transfers to that account fail, making `claim_slash` a dead end.

## Fix

Added `rescue_pending_slash` — an owner-callable method that redirects stuck funds to an alternative recipient, breaking the permanent lock cycle.

```rust
#[ink(message)]
pub fn rescue_pending_slash(&mut self, swap_id: u64, recipient: AccountId) -> Result<(), Error>
```

- Only contract owner can call (access controlled via `ensure_owner()`)
- Removes the pending slash entry and transfers funds to the specified recipient
- If the rescue transfer itself fails, re-inserts the pending slash (no fund loss)
- Emits `SlashClaimed` event for tracking

## Tests

Added 8 unit tests covering the full slash lifecycle:

- `test_timeout_slash_transfer_succeeds` — happy path timeout with successful transfer
- `test_timeout_slash_depletes_collateral_deactivates_miner` — collateral fully depleted
- `test_claim_slash_pays_out` — user successfully claims pending slash
- `test_claim_slash_wrong_user_rejected` — wrong caller can't claim
- `test_claim_slash_nonexistent` — claim on non-existent slash fails
- `test_owner_rescue_pending_slash` — owner redirects stuck funds
- `test_rescue_pending_slash_not_owner` — non-owner can't rescue
- `test_rescue_nonexistent` — rescue on non-existent slash fails

## Test plan

- [x] `cargo test` — all 8 tests pass
- [x] Verify on testnet that `rescue_pending_slash` correctly transfers funds to alternative recipient
- [x] Confirm only contract owner can call `rescue_pending_slash`

### DIFF
--- a/smart-contracts/ink/lib.rs
+++ b/smart-contracts/ink/lib.rs
@@ -926,6 +926,28 @@ mod allways_swap_manager {
             Ok(())
         }
 
+        /// Owner rescue for stuck pending slashes. If the original user's account
+        /// cannot receive transfers (destroyed, reaped, etc.), the owner can redirect
+        /// the payout to an alternative recipient to prevent permanent fund lock.
+        #[ink(message)]
+        pub fn rescue_pending_slash(&mut self, swap_id: u64, recipient: AccountId) -> Result<(), Error> {
+            self.ensure_owner()?;
+            let (_user, amount) = self.pending_slashes.get(swap_id).ok_or(Error::NoPendingSlash)?;
+
+            self.pending_slashes.remove(swap_id);
+            self.env().transfer(recipient, amount).map_err(|_| {
+                self.pending_slashes.insert(swap_id, &(_user, amount));
+                Error::TransferFailed
+            })?;
+
+            self.env().emit_event(SlashClaimed {
+                swap_id,
+                user: recipient,
+                amount,
+            });
+            Ok(())
+        }
+
         // =====================================================================
         // Miner Activation / Deactivation
         // =====================================================================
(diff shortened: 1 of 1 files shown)


LandynDev:
I appreciate the instinct to close a fund-flow edge case, but two things stacked against this landing.
                                                                                                                                                            
first, I believe that the "permanent lock" isn't actually permanent. a reaped substrate account isn't destroyed — it's just removed from storage below the existential deposit. the user holds the key; funding their AccountId with ED from any other wallet revives it, and claim_slash (lib.rs:887) works from there. i.e. recovery is user-driven, no contract change needed.                                                                                                                  
                                                          
second, and the bigger issue: `rescue_pending_slash` pulls in the opposite direction from where the contract is going. the trajectory is toward minimizing owner privilege. An owner-callable redirect for user funds has no
cryptographic basis for identifying the rightful recipient, which makes it a governance primitive, not a trustless one

PR #28, opened 2026-04-09
Clean up per-validator vote data on swap completion

dripsmvcp:
## Problem

During a security audit of the ink! smart contract, I found that per-validator vote entries are never cleaned up, creating dead storage that grows indefinitely.

**What happens:**

1. When validators vote on swaps, entries are written to `swap_confirm_votes[(swap_id, validator)]` and `swap_timeout_votes[(swap_id, validator)]`
2. When a swap completes or times out, only the aggregate `swap_confirm_vote_count` and `swap_timeout_vote_count` are removed (lines 758-759, 840-841)
3. The individual per-validator boolean entries are **never deleted** — they accumulate forever
4. Same issue in the request voting system: `request_votes[(request_id, validator)]` entries are never removed by `clear_request_data`

Over time this creates unbounded dead storage growth proportional to `num_swaps × num_validators`.

## Fix

Track voters per swap/request using `Vec<AccountId>` mappings, then iterate and remove all per-validator entries on completion:

- Added `swap_confirm_voters`, `swap_timeout_voters`, `request_voters` storage fields (`Mapping<u64, Vec<AccountId>>`)
- Vote recording now appends caller to the voters list
- New `clean_swap_votes` helper removes all per-validator entries for both confirm and timeout votes
- Updated `clear_request_data` to remove per-validator `request_votes` entries
- Both `vote_confirm` and `timeout_swap` now call `clean_swap_votes` at quorum instead of only removing vote counts

## Test plan

- [x] `cargo test` — `test_timeout_cleans_up_vote_data` confirms per-validator entries are removed after timeout
- [x] Verify on testnet that vote data is properly cleaned after swap completion

### DIFF
--- a/smart-contracts/ink/lib.rs
+++ b/smart-contracts/ink/lib.rs
@@ -40,8 +40,10 @@ mod allways_swap_manager {
         swaps: Mapping<u64, SwapData>,
         swap_confirm_votes: Mapping<(u64, AccountId), bool>,
         swap_confirm_vote_count: Mapping<u64, u32>,
+        swap_confirm_voters: Mapping<u64, Vec<AccountId>>,
         swap_timeout_votes: Mapping<(u64, AccountId), bool>,
         swap_timeout_vote_count: Mapping<u64, u32>,
+        swap_timeout_voters: Mapping<u64, Vec<AccountId>>,
         used_source_tx: Mapping<String, bool>,
 
         // Miner state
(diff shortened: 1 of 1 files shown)


LandynDev:
Thanks for this, at the time of your contribution we were landing a planned contract simplification (#63) that folded the four swap-lifecycle vote mappings into the universal request-id system (pending_swap_votes), with uniform cleanup on completion and timeout. Great contribution — clearly same vision.

PR #35, opened 2026-04-11
fix: use actual fee_divisor in swap receipt instead of hardcoded 100

Grizouforever:
## Summary

`_display_receipt` hardcodes `fee_divisor = 100` and displays `(1%)` while every other function in the swap flow reads the fee divisor from the contract via `get_fee_divisor()`. If the on-chain fee changes via `alw admin set-fee-divisor`, the receipt shows an incorrect fee amount and percentage.

## Changes

**`allways/cli/swap_commands/swap.py`**:
- Add `fee_divisor` parameter to `_display_receipt()` with `DEFAULT_FEE_DIVISOR` fallback
- Compute fee percentage dynamically instead of hardcoding `(1%)`
- Pass `fee_divisor` from the caller where it's already read from the contract

### DIFF
--- a/allways/cli/swap_commands/swap.py
+++ b/allways/cli/swap_commands/swap.py
@@ -277,7 +277,7 @@ def broadcast_reserve_with_retry(
 # =========================================================================
 
 
-def _display_receipt(swap):
+def _display_receipt(swap, fee_divisor: int = DEFAULT_FEE_DIVISOR):
     """Show a rich completion receipt after a successful swap."""
     src_chain_def = get_chain(swap.source_chain)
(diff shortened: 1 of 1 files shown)


Grizouforever:
Hi! This PR is ready for review — all 139/139 tests pass. Happy to address any feedback. Thanks!

LandynDev:
Thanks for this, at the time of your contribution we were landing a planned CLI + contract audit (#65). display_receipt's hardcoded fee_divisor = 100 was replaced with the FEE_DIVISOR constant as part of that pass. Great contribution with clearly same vision.

PR #36, opened 2026-04-11
skip extend reservation vote when reservation already expired

nudiltoys-cmyk:
## summary

I noticed `_try_extend_reservation` fires `vote_extend_reservation` even when the miner's reservation has already lapsed. the condition `reserved_until < current_block + EXTEND_THRESHOLD_BLOCKS` is true for any `reserved_until` behind `current_block`, so expired reservations trigger a contract call that gets rejected with `NoReservation`

changed the guard to `0 < blocks_left < EXTEND_THRESHOLD_BLOCKS` so it only fires when the reservation is still alive but within the threshold window

## test plan

verified the condition against blocks_left values of -5, 0, 1, 19, 20, 21, only 1 through 19 should trigger the extend vote

### DIFF
--- a/allways/validator/forward.py
+++ b/allways/validator/forward.py
@@ -78,7 +78,7 @@ def _try_extend_reservation(self: Validator, item, current_block: int, swap_labe
     try:
         reserved_until = self.contract_client.get_miner_reserved_until(item.miner_hotkey)
         blocks_left = reserved_until - current_block
-        if reserved_until < current_block + EXTEND_THRESHOLD_BLOCKS:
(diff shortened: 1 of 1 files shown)


LandynDev:
hey — appreciate the eye on this, but this one's been overtaken by the post-v1 refactor in #46 and #47 

- **already prevented upstream** — `state_store.purge_expired_pending_confirms()` at `forward.py:44` drops expired pending_confirms before `try_extend_reservation` is even reached, so the `blocks_left ≤ 0` path you're guarding shouldn't be hit in normal flow anymore.

PR #39, opened 2026-04-11
fix: return neutral speed score for miners with no timing data, remove dead all_chains

### LINKED ISSUE #38: Miners with completed swaps but no fulfillment timing receive zero reward
## Summary

When `fulfilled_block == 0` for a swap, no timing entry is added to `fulfillment_times_by_chain`. As a result `_chain_weighted_speed` iterates over an empty dict, `total_swaps` stays 0, and the function returns `0.0`. Because reward is computed as `success_rate * volume_weight * speed_score`, any miner whose completed swaps all lack timing data receives **zero reward** despite having fulfilled every swap.

Additionally, `all_chains: set = set()` (line ~385) is populated inside the loop but never read afterwards — dead code that misleads readers.

## Impact

- A miner that processes swaps where `fulfilled_block` is not written (e.g. data lag, off-chain settlement paths) is silently penalised to 0 reward, regardless of their actual success rate or volume.
- The dead `all_chains` variable inflates the code surface and implies a filtering step that does not occur.

## Expected Behaviour

`_chain_weighted_speed` should return a neutral baseline score (1.0) when the miner has completed swaps but none produced timing data, so the reward degrades gracefully to `success_rate * volume_weight` rather than collapsing to zero.

## Steps to Reproduce

1. Add a `COMPLETED` swap to `tracker.window` whose `fulfilled_block == 0`.
2. Call `calculate_miner_rewards`.
3. Observe the miner's reward is 0.0 even though `completed > 0`.

## Fix

See linked PR.

Grizouforever:
Closes #38

## Summary

- `_chain_weighted_speed` returned `0.0` when `total_swaps == 0`, which collapsed `success_rate * volume_weight * speed_score` to **zero** for any miner whose completed swaps all had `fulfilled_block == 0` (no timing recorded). Fix returns `1.0` (neutral baseline) when `total_swaps == 0` but `s.completed > 0`.
- Remove dead `all_chains: set` variable (populated in the loop, never consumed).

## Changes

**`allways/validator/forward.py`**:
- `_chain_weighted_speed`: add fallback `return 1.0` branch when no timing data exists but `s.completed > 0`
- `calculate_miner_rewards`: remove unused `all_chains` variable and its `all_chains.add(chain)` call

## Testing

All 139 existing tests pass (`uv run pytest`).

@anderdc @LandynDev

### DIFF
--- a/allways/validator/forward.py
+++ b/allways/validator/forward.py
@@ -382,11 +382,9 @@ def calculate_miner_rewards(
 
     # Per-chain fastest averages: compare miners only against others
     # fulfilling on the same dest chain (BTC vs BTC, TAO vs TAO, etc.)
-    all_chains: set = set()
     avg_speeds_by_chain: Dict[str, Dict[int, float]] = {}
     for uid, s in active_stats.items():
         for chain, times in s.fulfillment_times_by_chain.items():
-            all_chains.add(chain)
             if chain not in avg_speeds_by_chain:
                 avg_speeds_by_chain[chain] = {}
             avg_speeds_by_chain[chain][uid] = mean(times)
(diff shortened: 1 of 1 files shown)


LandynDev:
Thanks for this, at the time of your contribution we were landing a planned V1 incentive overhaul (#43) which replaced the speed-scoring path with a crown-time-per-direction model. The zero-reward edge case you caught disappears in the new model.

PR #41, opened 2026-04-11
fix: route all weight to recycle_uid when subnet price fetch fails

### LINKED ISSUE #40: Silent emission drop when subnet price fetch fails in apply_recycle
## Summary

`apply_recycle` in `allways/validator/recycle.py` (lines ~48-51) catches an exception from `get_subnet_price` and immediately returns `(rewards, uids)` — the unchanged inputs. When this happens during a normal scoring round, `_score_miners` calls `update_scores` with all-zero rewards and no `recycle_uid` entry. That round's full emission is silently discarded rather than recycled.

## Impact

Every transient RPC failure on the subnet price call causes one complete scoring interval's worth of emissions to vanish with no log entry indicating the loss. Over time (or during network instability) this accumulates into a meaningful, invisible emission leak.

## Expected Behaviour

When the subnet price is unavailable the safe fallback is to treat the situation identically to "no miners scored": route all weight to `recycle_uid` (`return np.array([1.0]), {recycle_uid}`). This matches the existing guard on lines 41-43 and ensures emissions are always either distributed to miners or recycled — never dropped.

## Steps to Reproduce

1. Mock `subtensor.get_subnet_price` to raise an exception.
2. Call `_score_miners` with a non-empty tracker window.
3. Observe `update_scores` is called with all-zero rewards and the `recycle_uid` is absent from the uid set.

## Fix

See linked PR.

Grizouforever:
Closes #40

## Summary

`apply_recycle` previously returned `(rewards, uids)` unchanged when `get_subnet_price` raised an exception. Because `rewards` at that point contains the raw miner scores (which sum to ≤ 1.0 but exclude `recycle_uid`), calling `update_scores` with those values silently drops the recycled portion for that round — emissions are neither paid to miners nor recycled.

The fix falls back to `(np.array([1.0]), {recycle_uid})`, identical to the "no miners scored" path, ensuring emissions are always recycled rather than lost.

## Changes

**`allways/validator/recycle.py`**:
- On `get_subnet_price` exception, return `([1.0], {recycle_uid})` instead of the unchanged `(rewards, uids)`
- Update the warning log message to describe what is actually happening

## Testing

All 139 existing tests pass (`uv run pytest`).

@anderdc @LandynDev

### DIFF
--- a/allways/validator/recycle.py
+++ b/allways/validator/recycle.py
@@ -47,8 +47,10 @@ def apply_recycle(
     try:
         alpha_price_tao = self.subtensor.get_subnet_price(self.config.netuid).tao
     except Exception as e:
-        bt.logging.warning(f'Recycle: failed to get subnet price, skipping: {e}')
-        return rewards, uids
+        bt.logging.warning(
+            f'Recycle: failed to get subnet price, routing all weight to recycle_uid to avoid silent emission drop: {e}'
+        )
+        return np.array([1.0], dtype=np.float32), {recycle_uid}
     window_fraction = SCORING_WINDOW_BLOCKS / 7200
     emission_alpha = DAILY_EMISSION_ALPHA * window_fraction
     emission_tao = emission_alpha * alpha_price_tao


LandynDev:
Thanks for this, at the time of your contribution we were landing a planned V1 incentive overhaul (#43) which removed the legacy scoring-window / fees / recycle paths entirely. The buggy "silent emission drop on price fetch failure" path no longer exists.

PR #34, opened 2026-04-11
fix: classify testnet P2SH addresses (prefix '2') as P2SH_P2WPKH

Grizouforever:
## Summary

Fixes #33

Testnet P2SH addresses (prefix `2`) are misclassified as `P2PKH` in `detect_address_type()`. On mainnet, prefix `3` correctly returns `P2SH_P2WPKH`, and `_to_mainnet_address` also correctly identifies `2` as P2SH (version byte `0xC4` → `0x05`).

## Changes

**`allways/chain_providers/bitcoin.py`**: Split the testnet condition so `2` returns `ADDR_TYPE_P2SH_P2WPKH` while `m`/`n` continue to return `ADDR_TYPE_P2PKH`.

## Impact

Fixes incorrect signing/verification for testnet P2SH-P2WPKH addresses and correct script type derivation in `send_amount_lightweight`.

PR #44, opened 2026-04-14
fix: correct testnet P2SH address classification bug in bitcoin provider

### LINKED ISSUE #33: Bug: Testnet P2SH addresses (prefix '2') misclassified as P2PKH
### Description

In `detect_address_type()`, testnet P2SH addresses (prefix `2`) are grouped with `m`/`n` prefixes and return `ADDR_TYPE_P2PKH`. On mainnet, prefix `3` correctly returns `ADDR_TYPE_P2SH_P2WPKH` (line 27), but the testnet equivalent `2` is misclassified.

The companion function `_to_mainnet_address` (line 58-61) correctly identifies `2`-prefix as P2SH (converts version byte `0xC4` to `0x05`), confirming the intended type.

### Impact

On testnet/regtest, signing or verifying messages for P2SH-P2WPKH addresses uses the wrong type, producing invalid signatures. Also affects `send_amount_lightweight` which derives the script type from `detect_address_type`.

### Fix

Split the condition so `2` returns `ADDR_TYPE_P2SH_P2WPKH` while `m`/`n` continue to return `ADDR_TYPE_P2PKH`.

Tmalone1250:
### Description
Corrected the Bitcoin address classification logic in `detect_address_type` to properly handle Testnet P2SH addresses.

### Technical Details
- Identified that addresses starting with prefix '2' (Testnet P2SH) were previously misclassified as `ADDR_TYPE_P2PKH` because they were grouped with 'm' and 'n' prefixes.
- Split the condition to explicitly return `ADDR_TYPE_P2SH_P2WPKH` for addresses starting with '2'.
- Verified the fix with a standalone logic test script across multiple address types (Mainnet P2PKH/P2SH, Testnet P2PKH/P2SH).

Fixes #33 
Reference: Sovereign Protocol local-first workflow.


LandynDev:
This work was completed by PR #34

PR #45, opened 2026-04-15
feat: Shell completion command

seroperson:
This PR introduces the `completion` command for enabling shell completion. With click [it's available out-of-box](https://click.palletsprojects.com/en/stable/shell-completion/), just needed a handy command to use it easily. Tools like `gh`, `pip`, `kubectl` etc have such `completion` command too, so it's a well-known adopted pattern.

```sh
eval "$(alw completion bash)" # bash
eval "$(alw completion zsh)"  # zsh
alw completion fish | source  # fish
```

https://github.com/user-attachments/assets/f5338ac1-41d1-4d40-952b-badaae77e91b

This aligns with the same functionality for `gitt`: https://github.com/entrius/gittensor/pull/431

The only thing which differs from `gitt` implementation is that `alw` uses heavy bittensor imports, which slow completions just a little, so I had to add a stub if we're inside of completion invocation. It's a common click limitation and it has several possible workarounds, but the most suitable is just stubbing.

### DIFF
--- a/allways/cli/main.py
+++ b/allways/cli/main.py
@@ -10,6 +10,8 @@ Usage:
     alw view                - View swaps, miners, rates
 """
 
+import os  # noqa: E402
+
 # Prevent bittensor from hijacking --help via its argparse config.
 # Must happen before any bittensor import.
 import sys as _sys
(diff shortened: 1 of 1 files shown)


LandynDev (approved):
Cool, thanks

PR #51, opened 2026-04-16
refactor: consolidate duplicate block-time and fee divisor constants

fansilas:
## Summary

Two numeric constants were defined in more than one place. This collapses each to a single source of truth without any behavior change.

- `SECONDS_PER_BLOCK = 12` in `allways/cli/swap_commands/helpers.py` duplicated `SUBTENSOR_BLOCK_SECONDS = 12` already defined in `allways/chains.py`. Eight CLI modules imported from `helpers`; they now import the canonical constant
  from `chains.py`.
- `fee_divisor: int = 100` was hardcoded as a default argument in `SwapFulfiller.__init__` and `SwapVerifier.__init__` while `FEE_DIVISOR = 100` already existed in `allways/constants.py`. Both defaults now reference the constant.
- `display_receipt` in `swap.py` had an inline `fee_divisor = 100  # 1% fee` that drifted from the same source. Replaced with `FEE_DIVISOR`, matching the existing pattern at `swap.py:634` and `quote.py:61`.

## Changes

- `allways/cli/swap_commands/helpers.py`: drop `SECONDS_PER_BLOCK`
- `allways/cli/swap_commands/{view,post_tx,swap,admin,status,resume,collateral}.py`: import `SUBTENSOR_BLOCK_SECONDS` from `allways.chains`, rename usages
- `allways/cli/swap_commands/swap.py`: `display_receipt` uses `FEE_DIVISOR` instead of the inline `100`
- `allways/miner/fulfillment.py`: `SwapFulfiller` default `fee_divisor = FEE_DIVISOR`
- `allways/validator/chain_verification.py`: `SwapVerifier` default `fee_divisor = FEE_DIVISOR`

## Testing

- All 267 tests pass (`pytest`)
- `ruff check allways/` clean
- `ruff format --check allways/` clean

No behavior change since `FEE_DIVISOR == 100` and `SUBTENSOR_BLOCK_SECONDS == 12`.
Runtime defaults verified: both `SwapFulfiller` and `SwapVerifier` still resolve `fee_divisor` to `100`.

### DIFF
--- a/allways/cli/swap_commands/admin.py
+++ b/allways/cli/swap_commands/admin.py
@@ -2,8 +2,9 @@
 
 import click
 
+from allways.chains import SUBTENSOR_BLOCK_SECONDS
 from allways.cli.help import StyledGroup
-from allways.cli.swap_commands.helpers import SECONDS_PER_BLOCK, console, from_rao, get_cli_context, loading, to_rao
+from allways.cli.swap_commands.helpers import console, from_rao, get_cli_context, loading, to_rao
 from allways.contract_client import ContractError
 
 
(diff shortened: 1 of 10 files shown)


LandynDev:
Thanks for this, at the time of your contribution we were landing a planned post-V1 cleanup (#46) which consolidated constants. FEE_DIVISOR is single-source in allways/constants.py and block-time constants are no longer duplicated. Great contribution with clearly same vision.

PR #52, opened 2026-04-16
refactor: centralize SCALE codec primitives in utils/scale.py

fansilas:
## Summary

SCALE primitive encoders and decoders were duplicated across three modules: `contract_client.py`, `validator/event_watcher.py`, and `validator/axon_handlers.py`.

Each copy carried its own magic byte widths, hex-prefix stripping, and compact-length encoders. This collapses them into a single `allways/utils/scale.py` that all three modules now import from.

No behavior change for well-formed contract or RPC data. Where the three copies disagreed on error handling (out-of-bounds `decode_string`, mode-2 compact address length, stray `0x` substrings), the extracted module picks the more defensive variant so malformed input fails gracefully rather than silently corrupting output.

## Changes

**New module: `allways/utils/scale.py`**

- Constants: `SS58_PREFIX`, `U32_BYTES`, `U64_BYTES`, `U128_BYTES`, `ACCOUNT_ID_BYTES`
- Helpers: `strip_hex_prefix`, `compact_encode_len`, `encode_bytes`, `encode_str`, `encode_u128`
- Streaming decoders returning `(value, new_offset)`: `decode_u32`, `decode_u64`, `decode_u128`, `decode_bool`, `decode_account_id`, `decode_string`

**`allways/contract_client.py`**

- Dropped the local `compact_encode_len` definition and inline hex-prefix / u128 / byte-width logic
- `encode_value`, `extract_*`, `decode_swap_data`, `get_reservation_data`, and `raw_contract_read` now delegate to the shared helpers
- Removed the one-line `decode_string` method that just forwarded to the utility
- Byte-width constants replace magic `4`, `8`, `16`, `32` literals

**`allways/validator/event_watcher.py`**

- Dropped the parallel copies of `decode_u32/u64/u128/bool/account_id/string` in favor of imports from `utils/scale`
- `topic_account_id` delegates to `decode_account_id`, and the hex-prefix strip in `to_bytes` uses `strip_hex_prefix`
- `struct` and `ss58_encode` imports removed (no longer needed locally)

**`allways/validator/axon_handlers.py`**

- The three `scale_encode_*_hash_input` functions now compose `encode_bytes`, `encode_str`, and `encode_u128` instead of inlining `compact_encode_len(len(...)) + ...` and `.to_bytes(16, 'little')`
- The local `encode_str` lambda inside `scale_encode_initiate_hash_input` is gone

**`tests/test_scale.py`**

- `TestDecodeString` exercises `decode_string` from the utility directly now that the client shim is gone
- `compact_encode_len` imported from its new home

## Testing

- All 267 tests pass (`pytest`)
- `ruff check allways/ tests/` clean
- `ruff format --check allways/ tests/` clean
- u128 encoding verified bit-identical against `struct.pack('<QQ', lo, hi)` across the range `[0, 2**128 - 1]` including boundaries at `2**64` and `2**127`
- String round-trip passes in all three SCALE compact-length modes (lengths 0, 5, 64, 16383, 16384)
- SS58 encoding via `ss58_encode(bytes, 42)` produces the same output as the previous `substrate.ss58_encode(bytes.hex())` path on Bittensor

### DIFF
--- a/allways/contract_client.py
+++ b/allways/contract_client.py
@@ -56,6 +56,22 @@ except ImportError:
 
 from allways.classes import Swap, SwapStatus
 from allways.constants import CONTRACT_ADDRESS, MIN_BALANCE_FOR_TX_RAO
+from allways.utils.scale import (
+    ACCOUNT_ID_BYTES,
+    U32_BYTES,
+    U64_BYTES,
+    U128_BYTES,
+    compact_encode_len,
+    decode_account_id,
+    decode_string,
+    decode_u32,
+    decode_u64,
+    decode_u128,
+    encode_bytes,
+    encode_str,
+    encode_u128,
+    strip_hex_prefix,
+)
 
 # =========================================================================
 # Contract selectors (from metadata — deterministic per contract build)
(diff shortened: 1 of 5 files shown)


LandynDev:
Hi @fansilas , thanks for the contribution. I think this could make the codebase cleaner from a quick skim, can you fix conflicts before I look more into it?

fansilas:
Hi @LandynDev 
Thanks for your review. I've fixed the conflict. Can you please review again?

LandynDev (approved):

PR #77, opened 2026-04-19
refactor: extract tao_leg_address helper for vote_initiate user_hotkey

### LINKED ISSUE #75: Extract tao_leg_address() helper to deduplicate user_tao_address derivation
## Context

The expression that picks the TAO-side SS58 address for `vote_initiate`'s `user_hotkey` is duplicated in two places:

- [allways/validator/forward.py:184](allways/validator/forward.py#L184)
- [allways/validator/axon_handlers.py:547](allways/validator/axon_handlers.py#L547)

Both compute the same thing:

```python
user_tao_address = to_address if to_chain == 'tao' else from_address
```

## Proposal

Extract a small helper into [allways/utils/rate.py](allways/utils/rate.py), co-located with the related `derive_tao_leg` (which mirrors the same TAO-side-of-the-swap concept for amounts):

```python
def tao_leg_address(from_chain, to_chain, from_address, to_address) -> str:
    return to_address if to_chain == 'tao' else from_address
```

Then replace the two inline expressions with calls to `tao_leg_address(...)`.

## Why

- Single source of truth for "which side of a swap is TAO" — pairs naturally with `derive_tao_leg` (amount) as the address counterpart.
- Keeps `vote_initiate` call sites consistent if the mapping ever changes (e.g. new TAO-equivalent chain).
- ~5 LOC addition, 2 call sites updated — trivial diff, no behavior change.

## Scope

- Add `tao_leg_address()` in `allways/utils/rate.py`.
- Update [allways/validator/forward.py:184](allways/validator/forward.py#L184).
- Update [allways/validator/axon_handlers.py:547](allways/validator/axon_handlers.py#L547).

dev-miro26:
Closes: #75 
## Summary
- Add `tao_leg_address(from_chain, to_chain, from_address, to_address)` in `allways/utils/rate.py`, co-located with `derive_tao_leg` (its amount counterpart).
- Replace the duplicated inline expression in `allways/validator/forward.py` and `allways/validator/axon_handlers.py` with calls to the helper.
- No behavior change — single source of truth for "which side of a swap is TAO" so the `vote_initiate` `user_hotkey` mapping stays consistent across call sites.

## Test plan
- [x] `ruff check allways/` clean
- [x] `pytest tests/` — 291 passed; 9 pre-existing failures in `tests/test_axon_handlers.py` unrelated to this refactor (verified on baseline)
- [x] Helper verified for both directions: `tao_leg_address('btc','tao',btc_addr,ss58)` → ss58; `tao_leg_address('tao','btc',ss58,btc_addr)` → ss58

### DIFF
--- a/allways/utils/rate.py
+++ b/allways/utils/rate.py
@@ -100,6 +100,11 @@ def derive_tao_leg(from_chain: str, from_amount: int, to_chain: str, to_amount: 
     return 0
 
 
+def tao_leg_address(from_chain: str, to_chain: str, from_address: str, to_address: str) -> str:
+    """Return the SS58 address on the TAO side of a swap — the user_hotkey for vote_initiate."""
+    return to_address if to_chain == 'tao' else from_address
+
+
 def check_swap_viability(
     tao_amount_rao: int,
     miner_collateral_rao: int,
--- a/allways/validator/axon_handlers.py
+++ b/allways/validator/axon_handlers.py
@@ -20,6 +20,7 @@ from allways.commitments import read_miner_commitment
 from allways.constants import RESERVATION_COOLDOWN_BLOCKS
 from allways.contract_client import AllwaysContractClient, ContractError, is_contract_rejection
 from allways.synapses import MinerActivateSynapse, SwapConfirmSynapse, SwapReserveSynapse
+from allways.utils.rate import tao_leg_address
 from allways.utils.scale import encode_bytes, encode_str, encode_u128
 from allways.validator.state_store import PendingConfirm
 
@@ -544,7 +545,9 @@ async def handle_swap_confirm(
             )
 
             # user_hotkey must be SS58 (TAO address): to_address for BTC→TAO, from_address for TAO→BTC
-            user_tao_address = synapse.to_address if swap_to_chain == 'tao' else synapse.from_address
+            user_tao_address = tao_leg_address(
+                swap_from_chain, swap_to_chain, synapse.from_address, synapse.to_address
+            )
             contract.vote_initiate(
                 wallet=validator.wallet,
                 request_hash=request_hash,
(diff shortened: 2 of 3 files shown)


anderdc:
Closing this along with #75.

**The abstraction isn't worth it at this scale, and the specific helper signature has a defect.**

**1. The helper has a dead parameter.**

```python
def tao_leg_address(from_chain: str, to_chain: str, from_address: str, to_address: str) -> str:
    return to_address if to_chain == 'tao' else from_address
```

`from_chain` is declared but never used. That's a real API bug — callers pass a value with zero effect, which misleads readers about what the function actually needs, makes unit tests awkward (any value for `from_chain` "works"), and invites silent argument-order mistakes because 3 of 4 string params look interchangeable to a caller.

**2. Abstracting a 1-line ternary into a 4-parameter function is over-engineering at 2 call sites.**

The original expression is 8 words on one line:

```python
user_tao_address = to_address if to_chain == 'tao' else from_address
```

A Python developer parses this in about a second. Replacing it with `tao_leg_address(from_chain, to_chain, from_address, to_address)` trades 90 characters for 107, adds a function hop, and doesn't compress the concept meaningfully.

**3. The existing inline comment is more informative than the helper name.**

At the axon handler call site:

```python
# user_hotkey must be SS58 (TAO address): to_address for BTC→TAO, from_address for TAO→BTC
user_tao_address = synapse.to_address if swap_to_chain == 'tao' else synapse.from_address
```

The comment explains **why** (SS58 is the TAO address format) and **which direction** (BTC→TAO vs TAO→BTC). The helper name `tao_leg_address` loses that context. Replacing the ternary with the helper either drops the comment (losing information) or keeps it (in which case the helper is redundant with the comment).

**4. "Centralizing the business rule" doesn't pay off here.**

The pitch is "if a new TAO-side chain is added, only one place needs updating." In practice, adding a new chain would also require changes to `derive_tao_leg`, the `vote_initiate` schema, chain-specific handlers, and the reservation logic. This helper doesn't materially ease that hypothetical change — it's still a multi-file edit.

**What would be a reasonable alternative**

If the duplication feels worth naming, add the existing axon-handler-side comment to the forward.py call site as well. That's zero new indirection, preserves the more-informative context, and documents the invariant.

Closing both the PR and the issue. Thanks for the diligence on duplication, just not every two-line pattern justifies extraction.

PR #80, opened 2026-04-19
refactor: extract _manage_validator to deduplicate add-vali and remove-vali commands

### LINKED ISSUE #79: Duplicated logic in add-vali and remove-vali CLI commands
## Description

`add_vali` and `remove_vali` in `allways/cli/swap_commands/admin.py` share an identical 30-line body. The only differences are the action word and which contract method is called.

Any future change — SS58 validation, dry-run flag, adjusted warning wording — must be applied to both functions independently, risking them falling out of sync.

## Location

`allways/cli/swap_commands/admin.py` — lines 294–363

wdeveloper16:
## Summary

- Extracted `_manage_validator(hotkey, *, adding)` shared helper to eliminate the duplicated body between `add-vali` and `remove-vali` commands
- Each command now delegates entirely to the helper via a single `adding=True/False` flag
- No behavior changes: all output strings, warnings, confirm prompts, and error messages are identical to the originals

## Changes

- `allways/cli/swap_commands/admin.py`: -42 / +26 lines
  - New `_manage_validator` helper handling the full flow: validator status check → conditional warning → confirm → submit transaction
  - `add_vali` and `remove_vali` reduced to one-line delegation calls

## Motivation

Both commands shared an identical 30-line skeleton. Any future change — e.g. adding SS58 validation, a dry-run flag, or adjusting the warning wording — would have required updating both functions independently.

## Test plan

- [ ] `alw admin add-vali <hotkey>` on an unregistered hotkey — no warning shown, confirms and submits
- [ ] `alw admin add-vali <hotkey>` on an already-registered hotkey — yellow warning shown before confirm
- [ ] `alw admin remove-vali <hotkey>` on a registered hotkey — no warning, confirms and submits
- [ ] `alw admin remove-vali <hotkey>` on an unregistered hotkey — yellow warning shown before confirm
- [ ] Cancelling either prompt prints `Cancelled` and exits without submitting
- [ ] RPC failure on `is_validator` prints error and exits before showing any UI

### DIFF
--- a/allways/cli/swap_commands/admin.py
+++ b/allways/cli/swap_commands/admin.py
@@ -291,6 +291,40 @@ def set_threshold(percent: int):
         print_contract_error('Failed to set consensus threshold', e)
 
 
+def _manage_validator(hotkey: str, *, adding: bool) -> None:
+    _, wallet, _, client = get_cli_context()
+    try:
+        is_registered = client.is_validator(hotkey)
+    except ContractError as e:
+        print_contract_error('Failed to check validator status', e)
+        return
+
+    action = 'Add' if adding else 'Remove'
+    console.print(f'\n[bold]{action} Validator[/bold]\n')
+    console.print(f'  Hotkey: {hotkey}')
+
+    if adding and is_registered:
+        console.print('  [yellow]Warning: This hotkey is already a registered validator[/yellow]')
+    elif not adding and not is_registered:
+        console.print('  [yellow]Warning: This hotkey is not a registered validator[/yellow]')
+
+    console.print()
+
+    if not click.confirm(f'Confirm {"adding" if adding else "removing"} validator?'):
+        console.print('[yellow]Cancelled[/yellow]')
+        return
+
+    try:
+        with loading('Submitting transaction...'):
+            if adding:
+                client.add_validator(wallet=wallet, validator=hotkey)
+            else:
+                client.remove_validator(wallet=wallet, validator=hotkey)
+        console.print(f'[green]Validator {hotkey} {"added" if adding else "removed"}[/green]\n')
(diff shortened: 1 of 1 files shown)


anderdc:
Closing — appreciate the attention to duplication, but the dedup here is marginal and mostly churn.

**Why:**

- **Semantically inverse operations forced through a direction flag.** `add_validator` and `remove_validator` aren't variants of one action — they're opposites. Merging them through `adding: bool` has the same readability cost as `operate_door(opening: bool)`: works, but obscures what operation is actually happening at each call site.
- **Five inline ternaries** on `adding` inside the helper encode the opposition (`"adding" if adding else "removing"`, `"added" if adding else "removed"`, etc.). Each one forces readers to re-parse which direction they're in.
- **Only 2 call sites.** Rule-of-three isn't dogma but at 2 inverse operations, the duplication cost is lower than the cognitive cost of the flag-parameterized helper.
- **Net -16 lines** is real but not compelling given the opacity tradeoff.

**If you wanted to dedupe something here**, the genuinely shared code is the preflight: `get_cli_context()` + `is_validator()` status check + console header + confirm prompt. That's roughly 10 lines common to both commands. The add/remove-specific bits (~5 lines each) are the inverse pieces that don't belong in a shared helper. A preflight-only extraction would net similar line savings without the ternary noise or flag-toggled logic.

No hard feelings — this is a judgment call and yours wasn't unreasonable. Just erring on the side of "leave inverse operations distinct" when the duplication is bounded to 2 call sites and the helper adds ternary density.

wdeveloper16:
Thanks for you review, @anderdc 
Good catch, that makes sense.
In the future, I will open meaningful PRs.

PR #92, opened 2026-04-19
fix: skip axon when is_validator RPC fails in discover_validators

### LINKED ISSUE #73: discover_validators can include non-whitelisted axons when is_validator RPC fails
**Problem**
In allways/cli/dendrite_lite.py, discover_validators optionally filters metagraph axons with contract_client.is_validator(hotkey). The is_validator call is wrapped in try/except Exception, and on any exception the code executes pass and then still appends the axon to the returned list.
That means a transient RPC / contract read failure behaves like “whitelist check passed” instead of “unknown — do not trust this endpoint”.

**Impact**
User-facing flows that rely on dendrite-lite (ephemeral wallet + broadcast) may send SwapReserve / SwapConfirm synapses to validator axons that are not on-chain whitelisted validators whenever is_validator errors. That weakens the intended gate and can waste user traffic, complicate debugging, or aid abusive endpoints during outages.

**Expected behaviour**
If the whitelist cannot be verified for a given hotkey, that axon should be skipped (fail closed), with a warning log — not included by default.

**Steps to reproduce (conceptual)**

1. Call discover_validators(subtensor, netuid, contract_client=client) with a contract_client whose is_validator raises (e.g. mock / inject RPC failure).
2. Observe axons for UIDs where is_validator would have returned False (or would have errored) are still present in the returned list.

**Proposed fix**

- On exception inside the whitelist branch: bt.logging.warning(...) and continue (or retry N times then skip).
- Optionally add a metric/log field with UID + hotkey for support.

**Acceptance criteria**

- When is_validator raises, the affected axon is not returned.
- When is_validator returns False, behaviour unchanged (still excluded).
- When is_validator returns True, behaviour unchanged (still included).
- Unit test with a mocked contract_client.is_validator that raises.

nightmare0329:
Fixes #73

An exception from contract_client.is_validator() was caught with pass, which fell through to axons.append() — including the axon as if the whitelist check had passed. Replace pass with continue so a failing RPC is treated as unknown (exclude) rather than whitelisted (include).

Changes:
- allways/cli/dendrite_lite.py — pass → continue in the is_validator exception handler

### DIFF
--- a/allways/cli/dendrite_lite.py
+++ b/allways/cli/dendrite_lite.py
@@ -62,7 +62,7 @@ def discover_validators(
                 if not contract_client.is_validator(metagraph.hotkeys[uid]):
                     continue
             except Exception:
-                pass
+                continue
         axons.append(axon)
 
     return axons


nightmare0329:
@LandynDev @FairyPigDev ready for review.

PR #49, opened 2026-04-16
refactor: remove dead read_result_option_swap and read_result_u128

t7929375-eng:
## Summary
Remove two dead methods from `AllwaysContractClient` in `allways/contract_client.py`:

- **`read_result_option_swap`** (line 730) — reads `Result<Option<SwapData>, ContractError>` return values. Never called anywhere in the codebase. Grep confirms zero references outside the definition.
- **`read_result_u128`** (line 752) — reads `Result<u128, ContractError>` return values. Never called anywhere in the codebase.

The codebase uses the non-Result-wrapped variants instead: `read_option_swap` (line 717) and `read_u128` (line 689).

## Related Issues
N/A (code cleanup)

## Type of Change
- [x] Refactoring (dead code removal)

## Testing
- [x] `ruff check` — 0 errors
- [x] `ruff format --check` — no diff
- [x] `pytest tests/ -v` — 267 passed, 0 failed

## Checklist
- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] No behavior change — removed code was unreachable

cc @anderdc @landyndev

PR #96, opened 2026-04-19
refactor: drop single-use ttl_cache wrapper, inline block cache in BaseNeuron

### LINKED ISSUE #95: Drop single-use ttl_cache wrapper in utils/misc.py
## Problem

`allways/utils/misc.py` carries a generic `@ttl_cache` decorator plus a `ttl_hash_gen` generator — roughly 30 lines of cache machinery whose only purpose is to back one function:

```python
@ttl_cache(maxsize=1, ttl=12)
def ttl_get_block(self) -> int:
    return self.subtensor.get_current_block()
```

Two things make this abstraction awkward:

1. **Single caller.** `ttl_get_block` is referenced exactly once in the codebase — [neurons/base/neuron.py:40](neurons/base/neuron.py#L40), inside `BaseNeuron.block`. Grep for `ttl_get_block | ttl_cache | ttl_hash_gen` returns hits in `utils/misc.py` itself and that one import line.
2. **Free-standing `self`.** `ttl_get_block(self)` takes `self` as an ordinary positional arg because `@ttl_cache` can't cleanly wrap bound methods — the decorator caches on the argument tuple, and passing a method through the decorator would need an `update_wrapper` dance that isn't there. So the function exists in `utils/misc.py` as a non-method that happens to accept a `BaseNeuron` instance, then is called from a class property. That's the abstraction leaking.

## Impact

- **Readability.** A reader tracing `BaseNeuron.block` has to cross a module boundary into `utils/misc.py`, follow the decorator through `ttl_cache` → `ttl_hash_gen` → `lru_cache`, and work out that the whole stack collapses to "cache the result for 12 seconds." That's five levels of indirection for a one-line cache.
- **Dead-flavored code.** `ttl_cache` and `ttl_hash_gen` are exported as if they're general-purpose utilities, but nothing else uses them. Anyone reusing them would be building on a pattern the codebase has otherwise avoided. Fits the tone of #49's "remove dead abstractions that pretend to be reusable."
- **Module sprawl.** `utils/misc.py` is the classic grab-bag — once the one useful thing is removed, there's no reason to keep the file.

## Proposed fix

Inline an instance-level cache into `BaseNeuron`:

```python
_BLOCK_CACHE_TTL_SECONDS = 12


class BaseNeuron(ABC):
    ...
    @property
    def block(self):
        """Current chain head, cached for one block (~12 s) to amortize the RPC round-trip."""
        now = time.monotonic()
        if now - self._cached_block_at >= _BLOCK_CACHE_TTL_SECONDS:
            self._cached_block = self.subtensor.get_current_block()
            self._cached_block_at = now
        return self._cached_block

    def __init__(self, config=None):
        ...
        self._cached_block: int = 0
        self._cached_block_at: float = 0.0
```

Then delete `allways/utils/misc.py` entirely — no callers remain.

`time.monotonic()` replaces the `time.time()` used inside `ttl_hash_gen`. This is strictly more robust (immune to wall-clock jumps, NTP slew, and leap seconds) and has identical semantics for a 12-second TTL — nothing in the codebase cares about wall vs. monotonic time at this granularity.

bitcompass:
Closes: #95 
## Summary
- Delete `allways/utils/misc.py` (30 lines of generic `ttl_cache` / `ttl_hash_gen` decorator machinery + the `ttl_get_block(self)` free function that sat on top of it).
- Replace with an instance-level cache on [`BaseNeuron`](neurons/base/neuron.py): `_cached_block` + `_cached_block_at`, guarded by the module-level `_BLOCK_CACHE_TTL_SECONDS = 12`.
- `BaseNeuron.block` is now a direct `time.monotonic()` check that calls `self.subtensor.get_current_block()` on miss — same 12-second window, same single call site.
- Drop the `from allways.utils.misc import ttl_get_block` import.

## Why
Mirrors the consolidation/dead-code tone of #49 and #71: the decorator was a generic abstraction built for one caller, and `ttl_get_block` took a free-standing `self` parameter purely so it could be decorated. A bound-method cache on the only class that needs it is simpler to read, removes a dangling import, and deletes the grab-bag module (`utils/misc.py`) altogether.

## Behavior note
The new cache uses `time.monotonic()` instead of `time.time()`. Strictly more robust (immune to wall-clock jumps, NTP slew, leap seconds) and identical semantics for the 12-second window — nothing in the codebase distinguishes wall from monotonic time at this granularity.

## Net
-47 lines (misc.py + 2 import lines + the old inline call site), +7 lines at `BaseNeuron.block` and `__init__`.

## Test plan
- [x] `ruff check allways/ neurons/` and `ruff format --check` — both clean.
- [x] `pytest tests/` — no new failures vs. current `test` branch (the 9 pre-existing `reservation_data`-mock failures are orthogonal and tracked separately).
- [x] Manual smoke: run a validator locally for one minute, confirm `self.block` keeps advancing and doesn't pin to a stale value (verify the TTL invalidation fires).
- [x] Grep confirms no remaining callers of `ttl_get_block` / `ttl_cache` / `ttl_hash_gen` / `allways.utils.misc`.


anderdc (approved):

bitcompass:
Hi, @anderdc 
Sorry bother you.
Could you please merge this PR?
Thank you.

bitcompass:
Hi, @anderdc
Sorry bother you.
Could you please merge this PR?
Thank you.

bitcompass:
Hi, @anderdc
Sorry bother you.
Could you please merge this PR?
Thank you.

bitcompass:
Hi, @LandynDev 
This PR was approved.
Could you please merge this PR?
Thanks

bitcompass:
Hi, @LandynDev
Could you please merge this PR?
Thanks

LandynDev:
- Trades a self-contained `ttl_cache` utility for two stateful instance attrs (`_cached_block`, `_cached_block_at`) baked into BaseNeuron — more class state, not less.
- Net -37 lines is misleading; the deleted module is a clean utility, the inlined version is just nested complexity in a base class.

PR #117, opened 2026-04-20
refactor: share reserve and swap proof message helpers

### LINKED ISSUE #116: Extract reserve and swap proof-message formats into a shared helper
The proof message strings that the CLI signs and the validator verifies are currently written as f-strings in both places:

* CLI signer at [allways/cli/swap_commands/swap.py:76](allways/cli/swap_commands/swap.py#L76): `proof_message = f'allways-swap:{from_tx_hash}'`
* CLI signer at [allways/cli/swap_commands/swap.py:164](allways/cli/swap_commands/swap.py#L164) and [allways/cli/swap_commands/swap.py:202](allways/cli/swap_commands/swap.py#L202): `reserve_proof_message = f'allways-reserve:{user_from_address}:{current_block}'`
* Validator verifier at [allways/validator/axon_handlers.py:282](allways/validator/axon_handlers.py#L282): `proof_message = f'allways-reserve:{synapse.from_address}:{synapse.block_anchor}'`
* Validator verifier at [allways/validator/axon_handlers.py:474](allways/validator/axon_handlers.py#L474): `proof_message = f'allways-swap:{synapse.from_tx_hash}'`

Any typo or format change on one side silently invalidates the proof on the other, with no structural guarantee that the two strings stay in lockstep. Moving these into small helpers (for example in `allways/synapses.py` or a new `allways/utils/proofs.py`):

```python
def reserve_proof_message(from_address: str, block_anchor: int) -> str:
    return f'allways-reserve:{from_address}:{block_anchor}'

def swap_proof_message(from_tx_hash: str) -> str:
    return f'allways-swap:{from_tx_hash}'

cleanjunc:
## Summary

Closes #116 

* Add allways/utils/proofs.py with reserve_proof_message and swap_proof_message as the single source of truth for the strings signed by the CLI and verified by the validator.
* Route the three CLI signer sites in allways/cli/swap_commands/swap.py and the two validator verifier sites in allways/validator/axon_handlers.py through the new helpers.
* Keep the BIP-137 roundtrip tests in tests/test_bitcoin_signing.py as is, since they exercise the Bitcoin provider rather than the proof format.

## Motivation
The reserve and swap proof formats were previously written as f-strings in both the CLI signer and the validator verifier. A typo or format change on one side would silently invalidate every proof on the other, with no structural guarantee that the two strings stayed in sync. Centralising the format removes that foot gun.

## Test plan
* [x] pytest tests/
* [x] ruff check allways/ neurons/
* [ ] Manual end to end: run a swap against a local validator and confirm both the reserve proof and the confirm proof verify.

### DIFF
--- a/allways/cli/swap_commands/swap.py
+++ b/allways/cli/swap_commands/swap.py
@@ -37,6 +37,7 @@ from allways.commitments import read_miner_commitments
 from allways.constants import FEE_DIVISOR, NETUID_FINNEY
 from allways.contract_client import ContractError
 from allways.synapses import SwapConfirmSynapse, SwapReserveSynapse
+from allways.utils.proofs import reserve_proof_message, swap_proof_message
 from allways.utils.rate import apply_fee_deduction, calculate_to_amount, check_swap_viability, derive_tao_leg
 
 
@@ -85,11 +86,10 @@ def sign_and_broadcast_confirm(
     """
     # Signing is fast for the internal path and may prompt interactively for
     # the external (paste-a-signature) path — never wrap it in a spinner.
-    proof_message = f'allways-swap:{from_tx_hash}'
     from_proof = sign_or_prompt_external(
         provider,
         user_from_address,
-        proof_message,
+        swap_proof_message(from_tx_hash),
         key=from_key,
         chain=from_chain,
         skip_confirm=skip_confirm,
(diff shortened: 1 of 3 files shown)


cleanjunc:
Hi @LandynDev @anderdc

LandynDev (dismissed):

LandynDev (approved):
Centralizes the reserve and swap proof-message formats into `allways/utils/proofs.py`, removing the foot-gun where the CLI signer and validator verifier each held their own f-string copy.

Verified after rebase onto current `test` (post-#119):
- All 4 inline f-string sites (2 CLI, 2 validator) now route through `reserve_proof_message` / `swap_proof_message`.
- No `allways-reserve:` / `allways-swap:` literals remain anywhere outside the helpers.
- Eliminated the local `reserve_proof_message` variable that would have shadowed the imported function.
- `ruff format` + `ruff check allways/ neurons/` clean.

Behavior is byte-for-byte identical to before; only the source of truth changes.

PR #121, opened 2026-04-20
fix: unify reserved_until boundary semantics across CLI and validator

### LINKED ISSUE #120: [Bug] Reserved-until boundary semantics differ across CLI, validator, and state store
`reserved_until` is compared against `current_block` in at least eight places, using three different semantics for the exact-boundary case where `reserved_until == current_block`:

* [allways/cli/swap_commands/post_tx.py:52](allways/cli/swap_commands/post_tx.py#L52) treats the boundary as expired: `if reserved_until <= current_block`.
* [allways/cli/swap_commands/resume.py:110](allways/cli/swap_commands/resume.py#L110), [allways/cli/swap_commands/status.py:80](allways/cli/swap_commands/status.py#L80), and [allways/cli/swap_commands/swap.py:246](allways/cli/swap_commands/swap.py#L246) also treat the boundary as expired (active iff `reserved_until > current_block`).
* [allways/cli/swap_commands/collateral.py:151](allways/cli/swap_commands/collateral.py#L151) treats the boundary as still reserved: `if reserved_until >= current_block`.
* [allways/validator/axon_handlers.py:335](allways/validator/axon_handlers.py#L335) treats the boundary as still reserved (`reserved_until >= validator.block`), and [allways/validator/axon_handlers.py:435](allways/validator/axon_handlers.py#L435) matches (`reserved_until < validator.block` for expired).
* [allways/validator/state_store.py:134](allways/validator/state_store.py#L134) keeps the pending confirm alive at the boundary: `DELETE FROM pending_confirms WHERE reserved_until < ?`.

At `current_block == reserved_until` a caller to `alw swap post-tx` is told the reservation has expired while the validator axon still accepts the confirm and the state store keeps the pending entry around for another block. The user-visible message and the on-chain path disagree for exactly one block per swap.

Folding the predicate into a single helper and calling it everywhere eliminates the divergence:

```python
def is_reserved(reserved_until: int, current_block: int) -> bool:
    return reserved_until > current_block
```

Once introduced, every callsite listed above reduces to `is_reserved(reserved_until, current_block)` (or its negation), and the state-store SQL purge aligns with it by using `reserved_until <= ?`. That way the CLI, the validator axon, and the persisted queue all agree on when a reservation lapses.

cleanjunc:
## Summary

Closes #120 

`reserved_until` was compared against `current_block` in ten places with three different boundary rules. At `current_block == reserved_until` the CLI reported the reservation expired, while the validator axon still accepted the confirm and the state store kept the pending entry alive for another block. The user visible message and the on chain path disagreed for exactly one block per swap.

The smart contract (`smart-contracts/ink/lib.rs`) is authoritative and treats `reserved_until >= current_block` as still reserved: `reserve` blocks new reservations while that predicate holds, and `initiate_swap` accepts the confirm at exactly the boundary. The CLI's strict `>` predicate was the outlier.

## Changes

* Added `is_reserved(reserved_until, current_block)` in `allways/utils/misc.py` returning `reserved_until >= current_block`, with a docstring pinning the semantic to the contract.
* Routed every Python call site through the helper: `post_tx`, `resume`, `status`, `swap`, `collateral`, `view`, and the two reservation gates in `validator/axon_handlers.py`.
* Kept the SQL purge in `state_store.py` as `reserved_until < ?` (the direct inverse of the helper) with a comment tying it back to `is_reserved`.

## Test plan

* [x] `pytest tests/test_pending_confirm_queue.py tests/test_axon_handlers.py::TestReservationValidation`
* [x] `ruff check .`


cleanjunc:
Hi @LandynDev @anderdc
Could you please review the PR? Thanks!

PR #122, opened 2026-04-20
fix: keep pending swap state until the swap appears on chain

### LINKED ISSUE #109: CLI deletes pending swap state before a queued swap is on-chain
### Problem
`pending_swap.json` is the CLI’s only saved recovery state. But it is deleted while the swap is still only pending:
- swap now deletes it on Ctrl+C and also when polling returns None, even while saying validators may still be processing the swap 
- swap resume does the same
- swap post-tx deletes it immediately when validators only queued the confirm and no swap ID exists yet

So the CLI drops recovery state before initiation is actually visible on-chain.

### Impact
This is real local data loss in a normal flow:

- user reserves a miner
- user sends funds and submits post-tx / resume
- validators accept but queue until confirmations land
- user exits, polling times out, or post-tx returns queued
- `pending_swap.json` is deleted before the swap exists on-chain

At that point the user loses the saved context needed for alw swap resume / alw swap post-tx. 
If the swap is still pending, delayed, or needs another retry, the normal recovery path is gone.

cleanjunc:
## Summary

Fixes #109 

* The CLI cleared `~/.allways/pending_swap.json` on Ctrl+C, polling timeout, and queued post-tx responses, wiping the only local recovery context while validators were still processing the swap.
* Extracts a shared `wait_for_swap_initiation` helper so the pending state is cleared only after a swap id is observed on chain. Ctrl+C and polling timeout now preserve the file so `alw swap resume`, `alw swap post-tx`, and `alw view reservation` still work afterwards.
* Drops the premature clear in the queued branch of `alw swap post-tx`, since no swap id exists at that point and validators may still take time to reach confirmation quorum.

## Test plan
* [x] Start `alw swap now`, press Ctrl+C during the "waiting for validators" phase, confirm the state file is still present and `alw swap resume` picks it up
* [x] Let the initial poll time out without a swap id, confirm the state file is preserved
* [x] Submit `alw swap post-tx` with a queued response from validators, confirm the file survives and `alw view reservation` still shows the pending context
* [x] Complete a swap end to end, confirm the state file is cleared once the swap id appears on chain
* [x] `ruff check allways/` and `pytest tests/` pass


cleanjunc:
Hi @LandynDev @anderdc
Could you please review the PR? Thanks!

anderdc:
#109's three data-loss paths are already fixed in the test branch — `swap now` and `resume` preserve pending_swap.json on Ctrl+C and poll timeout, and post-tx's queued branch has no clear. This diff is a behavior-neutral refactor, and it collides with #138 on the same poll-helper consolidation. Closing.

PR #124, opened 2026-04-20
fix: update queued reserved_until when reservation extends

### LINKED ISSUE #111: PendingConfirm can be dropped after ReservationExtended because local reserved_until is stale
### Problem
When `SwapConfirm` is queued, the validator stores a `PendingConfirm` with a snapshot of reserved_until.
Later, validators may successfully call `vote_extend_reservation()` and the contract updates the miner’s reservation expiry on-chain. But the queued `PendingConfirm` is not updated with the new expiry.

On each forward step, the validator runs `purge_expired_pending_confirms()` before replaying queued confirms. That purge uses the stale local `reserved_until`, not the current on-chain reservation expiry. 
As a result, a queued confirm can be deleted even though the reservation was validly extended on-chain.

### Impact
A valid queued `SwapConfirm` can be lost during normal operation after `ReservationExtended`.

Real effect:
- user submits SwapConfirm
- validator queues it because confirmations are still below threshold
- validators extend the reservation on-chain
- local queue still carries the old reserved_until
- purge deletes the queued PendingConfirm at the old expiry
- that validator stops auto-processing the still-valid queued confirm

This breaks the intended purpose of reservation extension for queued confirms and needs a patch.

cleanjunc:
## Summary

Closes #111 

Queued `PendingConfirm` entries kept a stale `reserved_until` from enqueue time. Once validators extended a miner's reservation on chain, the next run of `purge_expired_pending_confirms` would drop those entries as soon as the original expiry passed, silently abandoning swaps that should have been auto initiated.

## Fix

The validator event watcher now handles the `ReservationExtended` contract event and forwards the new `reserved_until` to a small state store helper. The helper issues one atomic `UPDATE ... WHERE reserved_until < ?`, which keeps the write monotonic (a replayed older event cannot shorten a fresher value) and a no op when no row exists for the miner.

## Changes

* `allways/validator/state_store.py`: adds `update_reserved_until`, an atomic monotonic bump on the pending_confirms row.
* `allways/validator/event_watcher.py`: routes `ReservationExtended` in `apply_event`, following the same type guard pattern as the other branches.
* `tests/test_pending_confirm_queue.py`: covers the bump path, the monotonic guarantee, and the missing miner no op.
* `tests/test_event_watcher.py`: end to end test that the event updates the queued entry.

## Testing

`uv run pytest tests/test_pending_confirm_queue.py tests/test_event_watcher.py` passes 25 tests, including 4 new ones added for this change.


cleanjunc:
Hi @LandynDev @anderdc

PR #125, opened 2026-04-20
fix: keep extended-timeout swaps seeded across validator restart

### LINKED ISSUE #110: Validator restart can permanently stop tracking extended active swaps
## Problem
`SwapTracker.initialize()` only seeds active swaps whose initiated_block is within one original fulfillment-timeout window of the current block. 
It stops scanning once:
```py
if swap.initiated_block < cutoff_block:
    break
```
But the contract supports `vote_extend_timeout()` and moves `swap.timeout_block` forward for still-active `FULFILLED` swaps. 
The validator itself uses that path in normal operation when destination confirmations are still pending near timeout.

So a swap can be older than the original `initiated_block` cutoff and still legitimately remain active because its timeout_block was extended.

## Impact
- A swap is `FULFILLED` but its destination tx still lacks enough confirmations.
- Validators use `vote_extend_timeout()` to keep it alive.
- A validator restarts after the swap becomes older than one original timeout window from `initiated_block`.
- That validator no longer tracks the still-active swap at all.

cleanjunc:
## Summary

Closes #110 

Restarting a validator silently stopped it from tracking any FULFILLED swap whose timeout had been extended beyond one fulfillment window from its initiation block. SwapTracker.initialize broke out of its backward scan on the first swap with `initiated_block < current_block - fulfillment_timeout_blocks`, so swaps kept alive by `vote_extend_timeout` vanished from the active set and the validator stopped verifying them or voting on their timeout.

The cold start now calls `get_active_swaps(max_gap=None)`, walking every stored ID from the latest down to 1. Because the contract prunes terminal swaps from storage, this returns exactly the live active set regardless of how many extensions a swap has accumulated. The default `max_gap` of 50 stays intact for CLI callers that prefer a fast, possibly stale view.

The `fulfillment_timeout_blocks` constructor parameter and the matching debug log fetch in `neurons/validator.py` were only consumed by the broken cutoff, so both are removed.

## Test Plan

* `pytest tests/test_swap_tracker.py` passes 26/26
* `pytest` passes 301/301
* `test_initialize_keeps_swap_with_extended_timeout` covers a FULFILLED swap whose `initiated_block` sits far behind but whose `timeout_block` has been extended forward
* `test_initialize_requests_full_scan` locks the `max_gap=None` contract so future changes cannot reintroduce an early break


cleanjunc:
Hi @LandynDev @anderdc

anderdc:
Issue #110's described mechanism doesn't exist — SwapTracker.initialize halts on consecutive pruned swap IDs (MAX_INIT_GAP), not on a block-age cutoff. The real change here replaces that bounded cold start with an unbounded next_id-length RPC scan at every validator restart, which regresses startup as the subnet accumulates swaps. Closing.

PR #129, opened 2026-04-20
Fix reserve-proof replay and raw-extrinsic sender off-by-one

### LINKED ISSUE #128: Two proof-handling bugs on the reservation path
## 1. Reserve proof is replayable

The message the user signs only binds `from_address` and `block_anchor`, and the validator never checks `block_anchor` against its current block:

```python
# allways/validator/axon_handlers.py:282
proof_message = f'allways-reserve:{synapse.from_address}:{synapse.block_anchor}'
```

It doesn't bind `miner_hotkey`, `from_chain`, `to_chain`, or any of the amounts. So one valid proof authorizes a reservation against **any miner, in any direction, for any amount**.

**Impact.** Reserve proofs travel in plaintext and leak easily (logs, past failed attempts, malicious validators). An attacker who grabs one valid proof for victim Alice can reserve any miner in her name repeatedly. Each expiry strikes Alice's address in the contract; cooldowns double per strike (150 → 300 → 600 → …). After ~5 replays Alice is locked out for days with no way to defend.

## 2. `parse_raw_extrinsic` reads sender bytes off by one

```python
# allways/chain_providers/subtensor.py:87
sender_bytes = body[1:33]
```

A v4 signed Substrate extrinsic is `[0x84][MultiAddress variant][32-byte AccountId]…`. Reading `body[1:33]` captures the variant byte plus only 31 of the 32 AccountId bytes, so the decoded SS58 is off by one byte on every extrinsic.

**Impact.** `ss58_encode` happily encodes any 32 bytes, so the error is silent — every raw-parsed block produces a plausible-looking but wrong sender. The raw path is a fallback for pruned/archival nodes; when it runs, `verify_transaction(expected_sender=…)` false-rejects legitimate TAO swaps and the sender-binding defense from PR #68 is eroded.

Bug has been present since the initial commit; no test exercises the raw parser against a real signed extrinsic.

wmagev:
Two proof-handling bugs on the reservation / transfer-verification path.

Reserve proof was signed over `allways-reserve:{from_address}:{block_anchor}` with no check that `block_anchor` is close to the validator's current block and no binding to miner_hotkey, chain direction, or amounts. An attacker who observes one valid proof could replay it against any miner, in any direction, for any amount; each expiry strikes the victim's address with a doubling cooldown and locks them out after a few replays.

- Add RESERVE_PROOF_MAX_AGE_BLOCKS freshness check against validator.block.
- Expand the signed message to bind miner_hotkey, from_chain, to_chain, tao_amount, from_amount, to_amount, block_anchor. Tag `v2` so pre-fix proofs stop verifying.
- Move the format into a shared `build_reserve_proof_message` helper so the CLI signer and validator verifier share one source of truth.

parse_raw_extrinsic in SubtensorProvider read `body[1:33]` to extract the sender, which captured the MultiAddress variant byte plus only 31 of the 32 AccountId bytes - every raw-parsed block produced a garbage SS58. Raw-parse is the fallback path for pruned/archival nodes; when it runs, expected-sender checks false-reject legitimate TAO swaps.

- Read `body[2:34]` after asserting the MultiAddress variant is Id (0x00); reject Index/Raw/Address{20,32} which the rest of the parser does not support.

CLI and validator must be upgraded together - old CLI against new validator will receive "Invalid source address proof" on every reservation.

Fixes #128

### DIFF
--- a/allways/chain_providers/subtensor.py
+++ b/allways/chain_providers/subtensor.py
@@ -81,10 +81,11 @@ class SubtensorProvider(ChainProvider):
             if not (body[0] & 0x80):
                 return None
 
-            # Sender AccountId is at bytes 1..33
-            if len(body) < 33:
+            # Signed v4 extrinsic: [0x84][MultiAddress variant][32-byte AccountId]...
+            # Only MultiAddress::Id (0x00) is supported; reject Index/Raw/Address{20,32}.
+            if len(body) < 34 or body[1] != 0x00:
                 return None
-            sender_bytes = body[1:33]
+            sender_bytes = body[2:34]
             sender = ss58_encode(sender_bytes, ss58_format=42)
 
             # Find the transfer call: pallet_index=5, call_index in {0,3,7}
--- a/allways/cli/swap_commands/swap.py
+++ b/allways/cli/swap_commands/swap.py
@@ -29,7 +29,7 @@ from allways.cli.swap_commands.helpers import (
 from allways.commitments import read_miner_commitments
 from allways.constants import FEE_DIVISOR, NETUID_FINNEY
 from allways.contract_client import ContractError
-from allways.synapses import SwapConfirmSynapse, SwapReserveSynapse
+from allways.synapses import SwapConfirmSynapse, SwapReserveSynapse, build_reserve_proof_message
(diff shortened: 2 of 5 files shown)


anderdc:
Fix conflicts.

PR #131, opened 2026-04-20
Gate halt on remaining vote paths and fail-close scoring check

### LINKED ISSUE #130: Halt lockdown is partially bypassed
## Contract skips halt on three voting paths

`smart-contracts/ink/lib.rs` — missing `ensure_not_halted()?` in:

- `vote_extend_reservation`
- `vote_initiate`
- `vote_extend_timeout`

During a halt, validators can still start new swaps and extend existing ones.

**Fix:** add `self.ensure_not_halted()?;` at the top of each function.

## Validator halt check fails open

`allways/validator/scoring.py:44-51` returns `False` on any exception, so an RPC hiccup makes the validator behave as if the contract is live.

**Fix:** fail closed — return `True` or skip the cycle.

wmagev:
Tracks two gaps that let the swap pipeline keep voting while the contract is halted: three ink! voting paths miss ensure_not_halted(), and the validator's halt check fails open on RPC errors.

Fixes #130

### DIFF
--- a/allways/validator/scoring.py
+++ b/allways/validator/scoring.py
@@ -42,13 +42,14 @@ def score_and_reward_miners(self: Validator) -> None:
 
 
 def _contract_is_halted(self: Validator) -> bool:
-    """Best-effort halt check. RPC flakiness should not zero every miner's
-    reward, so any exception falls through to normal scoring."""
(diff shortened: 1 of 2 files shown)


anderdc:
Fix conflicts.

PR #144, opened 2026-04-21
fix:treat ContractError as transient in SwapTracker.poll to prevent forward-step crash

edwin-rivera-dev:
## Summary

`SwapTracker.poll` catches known-transient network types (`ConnectionError`, `TimeoutError`, `asyncio.TimeoutError`) and re-raises anything else. `ContractError` did not appear in the transient list, so when `poll_inner` → `get_next_swap_id` → `read_u64` → `_read_typed` raised `ContractError('get_next_swap_id: no response')` on a subtensor blip, it matched the broad `except Exception` clause and got re-raised via the bare `raise`.

`forward.py:56` calls `await tracker.poll(self.block)` with no guard, and `forward()` has no outer try/except. The `ContractError` escapes all the way to base `run()`, which aborts the entire forward step and kicks in exponential backoff — skipping `confirm_miner_fulfillments`, `extend_fulfilled_near_timeout`, `enforce_swap_timeouts`, and the scoring gate for that tick.

This change adds `ContractError` to the transient-error tuple. `_read_typed` raises `ContractError` specifically when the underlying `raw_contract_read` returns `None`, and `poll_inner` is read-only (no business-logic rejections surface here), so a `ContractError` from this path is transient by construction.

## Change

- `allways/validator/swap_tracker.py` — import `ContractError`; add it to `poll`'s transient-error `except` tuple; docstring note explaining why it belongs there.

## Test plan

- [x] `ruff check allways/validator/swap_tracker.py`
- [x] `pytest tests/` — 279 passed
- [x] Behavioral smoke test:
  - Pre-fix: `ContractError` raised inside `poll_inner` escapes `poll` (bug reproduced)
  - Post-fix: `ContractError` caught in the transient clause, logged as warning, `poll` returns cleanly
  - Non-transient exception types (e.g. a fabricated `WeirdError`) still escalate via the unchanged `raise`


edwin-rivera-dev:
🐛 Fixed
Hi @anderdc , How are you? Could you review this PR please?
Thanks.

LandynDev:
ContractError from `get_next_swap_id` isn't realistically reachable — counter read, no revert path.

PR #151, opened 2026-04-21
CLI UX: clarify directions, add sort/filter, reshape view rates

LandynDev:
## Summary

- **`alw view rates` reshape** — user-shopping view: drop hotkey, add `Capacity (TAO)`, both-side addresses, a "do not send directly" warning, the contract `min_swap`/`max_swap` footer, and `--full` parity with `view miners`.
- **Sort + filter flags** — `--sort`, `--min-capacity`, `--search` on both `view miners` and `view rates`; plus `--status` on `view miners`.
- **Direction clarity pass** — unified legend (`BTC→TAO N reads 1 BTC → N TAO`) on both tables; `swap now` confirm panel uses explicit `You Send` / `You Receive` rows with a direction-aware `Rate` line; `swap quote` header frames direction and now prints the best miner's max-send capacity; `view swap`, `view reservation`, and `status` use `send X → receive Y` phrasing throughout.
- **`view miners` hotkey column** added (truncated unless `--full`).

Motivation: manual CLI vet turned up ambiguity in rate-column direction, missing info in `view rates` (no capacity, no send-to address, but an unhelpful hotkey), and no sort/filter affordance for a user scanning a long miner list.

## Test plan

- [x] `ruff format` + `ruff check` clean
- [x] `pytest tests/` — 300 pass
- [ ] Manual sweep against dev environment (`cd alw-utils/dev-environment && ./down.sh --force && ./up.sh --chains btc && alw view miners`, `alw view rates`, `alw view rates --sort capacity`, etc.)
- [ ] End-to-end swap via `alw swap now` on local dev env to confirm the new confirm-panel layout renders correctly


LandynDev:
Addressed both points:

**Rebased onto current `test`.** Force-pushed; diff is now CLI-only (4 files) — the lib.rs / metadata JSON residue from #149 is gone.

**Confirm-panel arrow ambiguity.** Rewrote the Swap Summary panel as two-row-per-side with explicit `From:` / `To:` labels (swap.py:713–722):

```
You Send:     <amount> <SRC>
  From:       <your SRC addr>  (your SRC address)

You Receive:  <amount> <DST>
  To:         <your DST addr>  (your DST address)
```

No amount-→-address arrow, so nothing to misread as a destination. The warning below the panel now references `"From" <SRC> address` to match.

While I was in there, caught one more instance of the same class of bug that predates this PR and is worth fixing together:

- `alw view swap` (build_swap_text) previously showed `Send to: {user_from_address}` — but `user_from_address` is the user's **source** address, i.e. where funds came **from**, not to. Relabeled to `From addr:` / `To addr:`.
- `alw view reservation` — the old labels `To (miner addr)` / `To (your addr)` were accurate but awkward and omitted the user's source address entirely. Rewrote as nested rows: `from (your SRC)`, `to (miner SRC)`, `to (your DST)`. The "Next step" line that prints when the reservation is active now says `Send N SRC from <your addr> to <miner addr>` instead of pointing vaguely at "the address above."

LandynDev:
Follow-up: removed `alw admin set-fee-divisor` as part of this PR.

The protocol fee is hardcoded at `FEE_DIVISOR=100` in both the ink! contract (`smart-contracts/ink/lib.rs:89`) and Python (`allways/constants.py:60`). There is no `set_fee_divisor` message on the contract and no `get_fee_divisor` / `set_fee_divisor` method on `ContractClient` — so the admin command would have crashed with `AttributeError` if anyone actually ran it. Pure dead code, dropped.

Docs don't reference it, so no doc-side change needed.

entrius (approved):

PR #153, opened 2026-04-21
fix: remove dead set-fee-divisor admin command

### LINKED ISSUE #152: Remove the broken `alw admin set-fee-divisor` command.
## Problem

The command currently calls `c.get_fee_divisor()` and `c.set_fee_divisor()` on the contract client, but neither method exists. Running the command always raises an `AttributeError`.

## Root Cause

This is not only a client-side issue. The smart contract defines the fee divisor as:

```rust
const FEE_DIVISOR: u128 = 100;

### PR DESCRIPTION
**Pull request by wdeveloper16**
## Summary

Removes the `alw admin set-fee-divisor` command, which was broken and could never work.

## Why

The command called `c.get_fee_divisor()` and `c.set_fee_divisor()` on the contract client — neither method exists, so running it always throws `AttributeError`.

More fundamentally, the smart contract defines the fee as:

```rust
const FEE_DIVISOR: u128 = 100;
```

A Rust `const` is compiled into the bytecode, not stored on-chain. There is no setter message in the contract and no way to change it without a full redeployment. The codebase already acknowledges this — `constants.py` has `FEE_DIVISOR = 100` with the comment *"Hardcoded 1% — matches the contract's immutable FEE_DIVISOR"*, and `alw view contract` displays the fee correctly by reading from that constant.

## What changed

- Deleted the 22-line `set_fee_divisor` command from `admin.py`
- No other changes — `alw view contract` fee display is unaffected

### DIFF
--- a/allways/cli/swap_commands/admin.py
+++ b/allways/cli/swap_commands/admin.py
@@ -87,29 +87,6 @@ def set_reservation_ttl(blocks: int):
     )
 
 
-@admin_group.command('set-fee-divisor', show_disclaimer=True)
-@click.argument('divisor', type=int)
-def set_fee_divisor(divisor: int):
-    """Set the fee divisor (100 = 1% fee, 50 = 2% fee, 20 = 5% fee max).
-
-    [dim]Examples:
-        $ alw admin set-fee-divisor 100[/dim]
-    """
-    if divisor < 20:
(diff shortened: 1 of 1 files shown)


wdeveloper16:
Hi, @LandynDev 

anderdc (approved):

PR #133, opened 2026-04-20
fix: validate src/dst addresses in alw miner post before chain commit

edwin-rivera-dev:
## Summary

Closes #132 

`alw miner post` accepted `src_addr` / `dst_addr` (from interactive prompts or positional args) without any format validation against their declared chains. A typo or cross-chain paste posted successfully and then silently broke every subsequent swap — reserve/confirm/send paths fail individually at the validator or miner, with no local indicator to the posting miner.

This change calls `provider.is_valid_address()` for both addresses against their respective chains before constructing the commitment string, mirroring the pattern used at [`allways/cli/swap_commands/swap.py:664-670`](allways/cli/swap_commands/swap.py#L664-L670) for user destination addresses.

## Change

- `allways/cli/swap_commands/pair.py`: import `create_chain_providers`; after `get_cli_context` resolves `subtensor`, instantiate chain providers and validate `src_addr` and `dst_addr` via `provider.is_valid_address`. Abort with a clear error on either failure.

## Behavior notes

- Both `BitcoinProvider.is_valid_address` and `SubtensorProvider.is_valid_address` are local (base58 / bech32 / SS58 format checks; no RPC). Validation is ~milliseconds.
- Validation runs after chain selection and after `get_cli_context` so the existing `subtensor` is reused — no extra connection.
- Addresses are validated against their user-declared chain, before the canonical-direction normalization. This matches user intent and keeps error messages keyed to what the user typed.
- No behavior change for already-valid postings.

## Test plan

- [x] `ruff check allways/`
- [x] `pytest tests/`
- [x] Manual: `alw miner post` with a good BTC + TAO pair — succeeds.
- [x] Manual: `alw miner post btc 5FsomethingSS58Looking tao bc1q...` — aborts with `Invalid BTC address: 5F...`.
- [x] Manual: `alw miner post tao 5F... btc bc1qTYPO` — aborts with `Invalid BTC address: bc1qTYPO`.
- [x] Manual: interactive wizard with a typo — aborts at the validation step before the confirmation prompt, giving the user a chance to retry without a pending commitment cost.

PR #156, opened 2026-04-21
fix: validate addresses in alw miner post before committing to chain

### LINKED ISSUE #155: `alw miner post` accepts invalid BTC/TAO addresses, risking stranded collateral and failed swap fulfillments
## Summary

`alw miner post` currently accepts arbitrary strings as BTC or TAO receiving addresses and commits them to the subnet without local validation.

If a miner submits a mistyped or malformed receiving address, matched swaps may later fail at fulfillment time because funds cannot be delivered to the configured address. This creates two problems:

- **Miner risk:** collateral can be put at risk because the miner cannot successfully receive funds for matched swaps
- **User impact:** users matched to that miner may experience failed fulfillments

## Impact

This is a correctness and safety issue in the miner setup flow.

Because invalid addresses are accepted at post time:
- configuration errors are persisted on-chain
- the mistake is only discovered much later, during fulfillment
- recovery is costly because swaps may already be matched against the miner

## Expected behavior

`alw miner post` should reject invalid BTC and TAO addresses before showing the final confirmation and before committing anything to the subnet.

Validation should:
- happen locally
- require no network calls
- produce a clear error message identifying the invalid chain/address
- stop execution immediately on the first invalid address

## Current behavior

Any string can be provided as a BTC or TAO address, and the command proceeds as though the addresses are valid.

## Reproduction

1. Run `alw miner post`
2. Provide a malformed BTC or TAO address (for example, a random string)
3. Observe that the command accepts it and continues through the posting flow
4. Once matched swaps rely on that address, fulfillment can fail

wdeveloper16:
## Summary

`alw miner post` accepted any string as a BTC or TAO address and committed it to the subnet without validation. A miner who typos their receiving address becomes unable to receive funds from matched swaps — their collateral is at risk and users matched to them will experience failed fulfillments.

## What changed

**`allways/cli/swap_commands/pair.py`**
- After chains and addresses are resolved (but before showing the confirmation summary), creates chain providers with `check=False` and calls `is_valid_address()` for both the source and destination address
- Both `BitcoinProvider` and `SubtensorProvider` validate locally — no network connections are made
- Rejects with a clear `[red]Invalid <ChainName> address: <addr>[/red]` message and returns early

**`tests/test_pair_validation.py`** (new, 4 tests)
- Invalid src address rejected
- Invalid dst address rejected  
- Valid addresses proceed to confirmation
- Both invalid shows only the first error

## Test Case

**Before**:

image

**After**:

image

### DIFF
--- a/allways/cli/swap_commands/pair.py
+++ b/allways/cli/swap_commands/pair.py
@@ -2,6 +2,7 @@
 
 import click
 
+from allways.chain_providers import create_chain_providers
 from allways.chains import SUPPORTED_CHAINS, canonical_pair
 from allways.cli.help import StyledCommand
 from allways.cli.swap_commands.helpers import console, get_cli_context, loading
@@ -149,6 +150,13 @@ def post_pair(
 
     config, wallet, subtensor, _ = get_cli_context(need_client=False)
     netuid = config['netuid']
+
+    providers = create_chain_providers(check=False, subtensor=subtensor, wallet=wallet)
+    for chain, addr in ((src_chain, src_addr), (dst_chain, dst_addr)):
+        p = providers.get(chain)
+        if p and not p.is_valid_address(addr):
+            console.print(f'[red]Invalid {SUPPORTED_CHAINS[chain].name} address: {addr}[/red]')
+            return
 
     rate_str = f'{rate:g}'
     counter_rate_str = f'{counter_rate:g}'


anderdc:
duplicate identified

wdeveloper16:
Hi, @anderdc 
Would you please clarify the closing reason more?

wdeveloper16:
I think it's really needed before moving fulfillment

wdeveloper16:
Hi, @anderdc
Could you review the PR, please?
My logic is smart and clear than #133.

PR #157, opened 2026-04-21
feat: implement save_state/load_state for miner swap poller cursor

wdeveloper16:
## Summary

Implements `save_state` / `load_state` on the `Miner` neuron to persist the swap poller cursor across restarts.

## Why

`BaseNeuron.sync()` calls `save_state()` after every forward iteration, but the miner's implementation was a no-op stub. On restart, `SwapPoller.last_scanned_id` reset to `0`, forcing an O(N) full re-scan of every swap ID in the contract before the miner could resume work. For a long-running subnet this scan grows unboundedly over time.

The miner's critical send-side state (`SwapFulfiller.sent_cache`) was already persisted to disk independently via JSON — so double-sends were never a risk. Only the polling cursor was being lost.

## What changed

**`neurons/miner.py`**
- `_state_path()` — returns `~/.allways/miner/state_{hotkey[:12]}.json`, consistent with the existing `sent_cache` path convention
- `save_state()` — writes `{"last_scanned_id": N}` to the state file; called automatically by `sync()` after each forward
- `load_state()` — reads the cursor back on startup; silently ignores missing file (fresh start) or corrupt JSON (logs a warning, leaves cursor at 0)
- `__init__` — calls `load_state()` after `swap_poller` is created so the first forward resumes from the saved position

## What's not changed

The validator's `save_state` / `load_state` (scores + hotkeys + step → NPZ) was already fully implemented in `BaseValidatorNeuron` — no changes needed there.

### DIFF
--- a/neurons/miner.py
+++ b/neurons/miner.py
@@ -96,6 +96,7 @@ class Miner(BaseMinerNeuron):
             pass
         except Exception as e:
             bt.logging.warning(f'Miner load_state failed, starting fresh: {e}')
+
     def unlock_coldkey(self) -> None:
         """Decrypt and cache the bittensor coldkey at startup so per-swap TAO
         transfers don't re-prompt for a password each time. Without this, every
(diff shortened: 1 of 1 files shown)


LandynDev:
Cold-start scan should use backward-scan + `MAX_INIT_GAP` pattern (validator's tracker) — no disk persistence needed.

PR #161, opened 2026-04-22
fix: extend swap scan window to include next_id-5 (off-by-one in range stop)

### LINKED ISSUE #150: swap: off-by-one in recent swap scan can hide real active swap
## Summary
The swap polling logic scans recent swap IDs using:

`for check_id in range(next_id - 1, max(next_id - 5, 0), -1):
`
Because Python range(..., stop, -1) excludes stop, this only checks 4 IDs (next_id-1 to next_id-4) and does not check next_id-5.

As a result, when the user’s swap is exactly 5 IDs behind next_id (or older), the CLI can fail to find it even though get_miner_has_active_swap(miner_hotkey) is true.

## Actual behavior
CLI may report swap not found/not initiated yet, even when an active swap exists on-chain.

## Expected behavior
If get_miner_has_active_swap(miner_hotkey) is true, polling should reliably locate the swap ID (or search a sufficiently large window / dedicated lookup path).

## Steps to reproduce
1. Have next_swap_id = N.
2. Ensure target swap exists at ID N-5 for the same miner_hotkey.
3. Run flow that calls polling (alw swap post-tx ... or alw swap now).
4. Observe polling misses this swap and can return None.

## Impact
False negative swap detection during confirmation flow, leading to misleading user output and unnecessary retries/manual checks.

Tet-9:
Closes #150

## Problem

The recent swap scan used:

```python
for check_id in range(next_id - 1, max(next_id - 5, 0), -1):
Python's range(start, stop, step) excludes the stop value. This means the loop only checked 4 IDs (next_id-1 through next_id-4) and never reached next_id-5. When a miner's active swap existed exactly at next_id-5 or older, the CLI would fail to find it and return None — even though get_miner_has_active_swap(miner_hotkey) confirmed an active swap exists on-chain.
This caused false negative swap detection during the confirmation flow, leading to misleading "swap not found / not initiated yet" output and unnecessary retries or manual checks.
Fix
Changed the range stop from next_id - 5 to next_id - 6 at both occurrences (lines 130 and 338):
for check_id in range(next_id - 1, max(next_id - 6, 0), -1):
This makes the range inclusive of next_id-5, scanning exactly 5 IDs as intended (next_id-1 through next_id-5).
Files Changed
allways/cli/swap_commands/swap.py — fixed off-by-one at both scan sites (lines 130 and 338)

### DIFF
--- a/allways/cli/swap_commands/swap.py
+++ b/allways/cli/swap_commands/swap.py
@@ -127,7 +127,7 @@ def poll_for_swap_creation(client, miner_hotkey: str) -> Optional[int]:
             try:
                 if client.get_miner_has_active_swap(miner_hotkey):
                     next_id = client.get_next_swap_id()
-                    for check_id in range(next_id - 1, max(next_id - 5, 0), -1):
+                    for check_id in range(next_id - 1, max(next_id - 6, 0), -1):
                         swap = client.get_swap(check_id)
                         if swap and swap.miner_hotkey == miner_hotkey:
                             return check_id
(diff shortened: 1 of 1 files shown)


Tet-9:
@anderdc , I need to know why you closed this without a merged commit? 
Everything was done as perfectly as it could have

PR #165, opened 2026-04-23
swap poller: use get_miner_active_swaps over bounded scan

### LINKED ISSUE #150: swap: off-by-one in recent swap scan can hide real active swap

LandynDev:
Closes #150.

The old `resolve_recent_swap_id` scanned a fixed 4-ID window back from `next_id-1`. Correct diagnosis in #150 (and PR #161), but widening the window to 5/100/N is still guessing.

Swaps it for the existing `client.get_miner_active_swaps(hotkey)` helper, which uses the standard 50-consecutive-nones scan pattern from `contract_client`. Also removes a duplicate inline scan in `poll_for_swap_with_progress`.

Net: -5 lines, no magic number, correct regardless of swap position.

### DIFF
--- a/allways/cli/swap_commands/swap.py
+++ b/allways/cli/swap_commands/swap.py
@@ -125,22 +125,20 @@ def sign_and_broadcast_confirm(
 
 
 def resolve_recent_swap_id(client, miner_hotkey: str) -> Optional[int]:
-    """Return the miner's most recent swap id, or None if they have none.
+    """Return the miner's active swap id, or None if they have none.
 
-    Walks back 4 IDs from ``next_id - 1`` as a defensive buffer against
-    concurrent swap creation. Raises ``ContractError`` on RPC failure so
-    callers can distinguish "RPC broken" from "no swap yet" — the
-    Ctrl+C-exit path suppresses; ``poll_for_swap_creation`` counts for
-    its retry warning.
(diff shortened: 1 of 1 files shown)


LandynDev:
@anderdc check this out

anderdc (approved):

PR #101, opened 2026-04-19
refactor: add is_already_voted predicate for AlreadyVoted rejections

dev-miro26:
Closes: #100
## Summary
- Add `is_already_voted(e)` alongside `is_contract_rejection` in `allways/contract_client.py` — a single source of truth for detecting the `AlreadyVoted` contract-error variant, anchored to `CONTRACT_ERROR_VARIANTS[3]`.
- Replace two inline `'AlreadyVoted' in str(e)` substring checks in `allways/validator/forward.py` with calls to the new predicate.
- No behavior change — the detection semantics are identical today; this is purely a centralization so the substring literal lives in exactly one place.

## Why
`forward.py:254` and `forward.py:395` were branching on `if 'AlreadyVoted' in str(e):`. The codebase already exports a structured `CONTRACT_ERROR_VARIANTS` map and a canonical `is_contract_rejection(e)` predicate from `allways/contract_client.py`, but the `AlreadyVoted` check bypassed both with a raw substring match. If the `ContractError` message format ever changes (e.g. variant rename, localisation, format tweak in `decode_contract_error`), the two call sites would silently stop matching with no lint or test signal.

Co-locating the predicate with `is_contract_rejection` also leaves room for a small documented predicate set (`is_duplicate_source_tx`, `is_miner_reserved`, …) if future code wants variant-specific branching.

## Changes
- **`allways/contract_client.py`:** add `is_already_voted(e) -> bool` next to `is_contract_rejection`; docstring points at `CONTRACT_ERROR_VARIANTS[3]` as the canonical variant source.
- **`allways/validator/forward.py`:**
  - Import `is_already_voted` alongside the existing `is_contract_rejection`.
  - `forward.py:254` (extend-reservation vote failure branch) — swap substring check for predicate.
  - `forward.py:395` (extend-timeout vote failure branch) — swap substring check for predicate.
- Net `+8 / −2` lines; no signature changes.

## Type of Change
- Refactoring (centralisation of an existing predicate, no behavior change)

## Test plan
- [x] `ruff check allways/ neurons/ tests/` — clean
- [x] `ruff format --check allways/ neurons/ tests/` — clean
- [x] `pytest tests/` — **291 passed**; 9 pre-existing failures in `tests/test_axon_handlers.py` on `upstream/test` are unrelated (`"too many values to unpack (expected 3)"` in the `SwapConfirm` handler) and unchanged by this diff
[...]

PR #180, opened 2026-04-24
Fix/centralize already voted predicate

### LINKED ISSUE #172: Brittle `'AlreadyVoted' in str(e)` substring check still present after PR #101
PR #101 introduced an `is_already_voted` predicate, but [allways/validator/forward.py:256](allways/validator/forward.py#L256) and [allways/validator/forward.py:397](allways/validator/forward.py#L397) still inline `if 'AlreadyVoted' in str(e):`. Whatever helper landed (or was meant to) is not used here. Either route both call sites through the centralized predicate or — if the helper was never merged — re-open it; the substring form silently breaks if the contract error type is renamed.

thomascolden585-svg:
## Summary
- Replace fragile inline checks (`'AlreadyVoted' in str(e)`) with a centralized predicate.
- Add `is_already_voted(e)` in `allways/contract_client.py` as the single source of truth for this contract rejection case.
- Route both validator forward-path call sites through the helper:
  - reservation extension handling in `try_extend_reservation`
  - timeout extension handling in `extend_fulfilled_near_timeout`
  - 
Closes #172

## Problem
`forward.py` still used inline substring checks for `"AlreadyVoted"` at two locations after PR #101 introduced a predicate approach. This created drift and risk: contract error naming changes could silently break these branches.

## Solution
- Implement and use `is_already_voted(e)` from `contract_client` in all affected paths.
- Keep rejection semantics consistent with existing centralized helpers like `is_contract_rejection(e)`.

## Why this is safer
- One place to maintain the "already voted" contract error contract.
- Avoids duplicated ad-hoc string matching in business logic.
- Reduces chance of silent regressions if contract error formatting changes.

## Files Changed
- `allways/contract_client.py`
- `allways/validator/forward.py`

## Validation
- Verified updated call sites now use `is_already_voted(e)`.
- Confirmed no remaining inline `"AlreadyVoted" in str(e)` checks in repository.
- Lint diagnostics for edited files reported no errors.
## Summary

- Replaced string-literal event routing in `allways/validator/event_watcher.py` with a shared `ContractEventName` enum to avoid silent no-op dispatch when labels drift.
- Added `allways/validator/contract_events.py` as the canonical source of contract event labels used by validator-side event handling.
- Hardened decode/dispatch flow so unknown metadata event labels are explicitly detected and warned on, instead of falling through default behavior.

## Why

Event routing previously depended on raw string comparisons (e.g. `'MinerActivated'`, `'SwapInitiated'`). If contract metadata labels change, decoded events could silently bypass intended branches. Centralizing labels in an imported enum makes the routing contract explicit and reduces drift risk.

## Changes

- **New module:** `allways/validator/contract_events.py`
  - Introduces `ContractEventName(str, Enum)` with:
    - `MINER_ACTIVATED = "MinerActivated"`
    - `SWAP_INITIATED = "SwapInitiated"`
    - `SWAP_COMPLETED = "SwapCompleted"`
    - `SWAP_TIMED_OUT = "SwapTimedOut"`

- **Updated:** `allways/validator/event_watcher.py`
  - Imports `ContractEventName` from the new module.
  - `decode_contract_event(...)` now validates metadata event labels by coercing to `ContractEventName`.
  - Unsupported metadata labels are logged and dropped explicitly.
  - `apply_event(...)` accepts `str | ContractEventName`, normalizes to enum, and warns on unsupported names.
  - Branch dispatch now uses enum members rather than string literals.

## Risk / Compatibility

- Low risk: behavior for known event labels is unchanged.
- Safer failure mode: unexpected labels now warn and are ignored explicitly instead of silently missing a string match.

## Test Plan

- [x] `python3 -m compileall allways/validator/event_watcher.py allways/validator/contract_events.py`
- [x] Lint check for touched files (`event_watcher.py`, `contract_events.py`) reports no new diagnostics.
- [ ] Run validator tests locally once `pytest` is available:
  - `pytest -q tests/test_event_watcher.py tests/test_scoring_v1.py`

### DIFF
--- a/allways/contract_client.py
+++ b/allways/contract_client.py
@@ -293,6 +293,17 @@ def is_contract_rejection(e: BaseException) -> bool:
     return 'contract rejected' in msg or 'ContractReverted' in msg
 
 
+def is_already_voted(e: BaseException) -> bool:
+    """Return True if ``e`` is the contract's "already voted" rejection.
+
+    This keeps all "already voted" matching in one place so call sites do not
+    hardcode error names or re-implement fragile string checks.
+    """
+    msg = str(e)
+    already_voted_name = CONTRACT_ERROR_VARIANTS.get(3, ('AlreadyVoted', ''))[0]
+    return f': {already_voted_name} ' in msg or f': {already_voted_name} —' in msg
+
+
 # =========================================================================
 # Client
 # =========================================================================
--- a/allways/validator/contract_events.py
+++ b/allways/validator/contract_events.py
@@ -1,0 +1,12 @@
+"""Canonical contract event labels consumed by the validator."""
+
+from __future__ import annotations
+
+from enum import Enum
+
+
+class ContractEventName(str, Enum):
+    MINER_ACTIVATED = 'MinerActivated'
+    SWAP_INITIATED = 'SwapInitiated'
+    SWAP_COMPLETED = 'SwapCompleted'
+    SWAP_TIMED_OUT = 'SwapTimedOut'
--- a/allways/validator/event_watcher.py
+++ b/allways/validator/event_watcher.py
@@ -30,6 +30,7 @@ from allways.utils.scale import (
     decode_u128,
     strip_hex_prefix,
 )
+from allways.validator.contract_events import ContractEventName
 from allways.validator.state_store import ValidatorStateStore
 
 DATA_DECODERS = {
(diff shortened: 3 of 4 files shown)


thomascolden585-svg:
HI, @LandynDev 
I have resolved the requested changes and updated the PRs accordingly.I would appreciate it if you could review them when convenient.
Thank you for your time.

PR #190, opened 2026-04-24
fix(event_watcher): halt cursor at first failed block

### LINKED ISSUE #175: `event_watcher.sync_to` advances cursor past failed blocks
`process_block` ([allways/validator/event_watcher.py:324-332](allways/validator/event_watcher.py#L324-L332)) catches all exceptions and returns silently, then `sync_to` unconditionally sets `cursor = end` at [:321](allways/validator/event_watcher.py#L321).

A transient WebSocket hiccup on a single block → `SwapCompleted` / `SwapTimedOut` events permanently dropped → affected miner stays marked busy, scoring goes wrong with no recovery path.

**Direction:** track per-block success and don't advance the cursor past blocks that threw.

dripsmvcp:
## Summary

Fixes entrius/allways#175 — `event_watcher.sync_to` advances cursor past failed blocks.

`sync_to` set `cursor = end` unconditionally, even when `process_block` caught an exception mid-loop. A transient RPC hiccup on one block dropped its `SwapCompleted` / `SwapTimedOut` events forever, leaving the affected miner marked busy with no recovery path.

## Change

- `process_block` now returns `bool` — `True` on success, `False` on exception or unavailable block hash.
- `sync_to` stops advancing at the last successful block. The next tick retries the failing one.
- A `None` / empty `block_hash` (pruned or not-yet-synced) also halts the cursor instead of being silently skipped.

## Trade-off

A persistently unreadable block will now stall the cursor rather than silently skip. That's intentional — surface the problem loudly so operators can investigate, rather than letting events go missing unnoticed. `MAX_BLOCKS_PER_SYNC` chunking still bounds how far `sync_to` falls behind per call.

## Test plan

- [x] `pytest tests/test_event_watcher.py` — 19/19 pass (16 prior + 3 new)
- New cases under `TestSyncToCursorAdvance`:
  - `test_cursor_stops_at_first_failed_block`
  - `test_retry_on_next_tick_replays_failed_block`
  - `test_none_block_hash_halts_cursor`

### DIFF
--- a/allways/validator/event_watcher.py
+++ b/allways/validator/event_watcher.py
@@ -312,24 +312,38 @@ class ContractEventWatcher:
 
     def sync_to(self, current_block: int) -> None:
         """Catch up from cursor to ``current_block`` in MAX_BLOCKS_PER_SYNC
-        chunks so a long outage doesn't freeze the forward loop."""
+        chunks so a long outage doesn't freeze the forward loop.
+
+        Stops advancing the cursor at the first block whose events we could
+        not fetch — a transient RPC hiccup would otherwise drop SwapCompleted
+        / SwapTimedOut events forever and strand miners in busy state. The
+        next tick will retry from the same point.
+        """
         if current_block <= self.cursor:
             return
         end = min(current_block, self.cursor + MAX_BLOCKS_PER_SYNC)
+        last_success = self.cursor
         for block_num in range(self.cursor + 1, end + 1):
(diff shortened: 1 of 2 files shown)


dripsmvcp:

LandynDev:
Hey, rebase asap please @dripsmvcp

LandynDev:
We will continue with your identical contribution: https://github.com/entrius/allways/pull/212

PR #191, opened 2026-04-24
fix(event_watcher): make SwapInitiated replay idempotent against bootstrap

### LINKED ISSUE #176: `apply_busy_delta` / `apply_event` not idempotent on restart replay
If the validator crashes mid-block, the same block is reprocessed on restart ([allways/validator/event_watcher.py:381-413](allways/validator/event_watcher.py#L381-L413)).

`insert_swap_outcome` uses `INSERT OR REPLACE` (fine), but the busy-delta increments double — each replayed event bumps the counter again.

**Direction:** dedupe by `(block, extrinsic_index, event_index)`, or store an absolute busy count instead of applying deltas.

dripsmvcp:
## Summary

Fixes entrius/allways#176 — `apply_busy_delta` / `apply_event` not idempotent on restart replay.

The issue frames this as a mid-block-crash scenario, but the bug fires on **every normal restart** where a live swap's `initiated_block` falls within the scoring window:

1. `initialize()` reads the contract's active-swap list and seeds `open_swap_count[hk] += 1` for each in-flight swap.
2. `initialize()` then rewinds the cursor to `current_block - SCORING_WINDOW_BLOCKS`.
3. `sync_to` replays events in the window. When it hits the original `SwapInitiated` for the same swap, `apply_busy_delta(+1)` fires again → `open_swap_count[hk]` becomes 2.

The `max(0, current + delta)` guard in `apply_busy_delta` silently masks the symptom for `-1` events, but a phantom `+1` leaks into scoring and the miner stays marked busy at count=2 forever (or until the next terminal event, which only brings it back to 1).

## Change

- New field `ContractEventWatcher.bootstrapped_swap_ids: Set[int]`.
- `initialize()` records each in-flight swap's `id` when seeding `open_swap_count` from the contract.
- `apply_event` for `SwapInitiated` now reads `swap_id` (previously discarded) and skips the `+1` delta if the id is in `bootstrapped_swap_ids`.
- `SwapCompleted` / `SwapTimedOut` handlers discard the id from the set after applying the `-1`, so the cleanup is tidy and the set doesn't grow unbounded across the scoring window.

```python
elif name == 'SwapInitiated':
    swap_id = values.get('swap_id')
    miner = values.get('miner', '')
    if miner:
        if isinstance(swap_id, int) and swap_id in self.bootstrapped_swap_ids:
            return  # +1 already seeded from contract; avoid double-count
        self.apply_busy_delta(block_num, miner, +1)

### DIFF
--- a/allways/validator/event_watcher.py
+++ b/allways/validator/event_watcher.py
@@ -226,6 +226,11 @@ class ContractEventWatcher:
         self.busy_events: List[BusyEvent] = []
         self.active_events: List[ActiveEvent] = []
         self.active_events_by_hotkey: Dict[str, List[ActiveEvent]] = {}
+        # Swap IDs whose +1 was seeded directly from the contract's active-swap
+        # list during initialize(). Replay must skip their SwapInitiated event
+        # to avoid double-counting — the busy tick is already in open_swap_count.
+        # Entries are discarded on the matching terminal event.
+        self.bootstrapped_swap_ids: Set[int] = set()
 
     # ─── Public API consumed by scoring ─────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


dripsmvcp:

LandynDev (approved):
Rebased + tested locally (411 passed, lint clean).

PR #192, opened 2026-04-24
fix(forward): re-raise non-provider errors in confirm_miner_fulfillments

### LINKED ISSUE #178: `asyncio.gather` swallows programming errors
[allways/validator/forward.py:331-334](allways/validator/forward.py#L331-L334) uses `return_exceptions=True`, so any exception from a verifier coroutine becomes a "result."

A typo in the verifier (e.g. `AttributeError`) gets logged as a verification error; the forward loop keeps running with silently-missing votes, masking real bugs as flaky network.

**Direction:** catch only provider/timeout exceptions; re-raise everything else.

dripsmvcp:
## Summary

Fixes entrius/allways#178 — `asyncio.gather` in `confirm_miner_fulfillments` swallows programming errors.

`asyncio.gather(..., return_exceptions=True)` turns every exception into a result. The old dispatch caught `ProviderUnreachableError` specifically, then used a broad `isinstance(result, Exception)` fallback that logged any other exception as "verification error" and continued the loop. A typo in the verifier (`AttributeError`, `KeyError`, etc.) was therefore silent — every affected swap skipped its vote and rolled to timeout, slashing honest miners for what looked like validator flakiness.

## Change

Replace the broad fallback with a re-raise for anything outside the known-transient set (currently just `ProviderUnreachableError`). A programming error now crashes the forward loop loudly on the first hit — strictly preferable to silent mis-scoring.

```python
if isinstance(result, ProviderUnreachableError):
    uncertain.add(swap.id)
    continue
if isinstance(result, BaseException):
    raise result              # ← was: log + continue

### DIFF
--- a/allways/validator/forward.py
+++ b/allways/validator/forward.py
@@ -337,9 +337,12 @@ async def confirm_miner_fulfillments(
             bt.logging.warning(f'Swap {swap.id}: provider unreachable, deferring verification')
             uncertain.add(swap.id)
             continue
-        if isinstance(result, Exception):
-            bt.logging.error(f'Swap {swap.id}: verification error: {result}')
-            continue
+        if isinstance(result, BaseException):
+            # Any other exception is a programming error, not a transient
+            # network failure. Re-raise so the forward loop crashes loudly
+            # rather than silently skipping votes and letting honest miners
+            # roll to timeout.
+            raise result
         if result:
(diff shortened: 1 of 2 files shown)


dripsmvcp:

LandynDev:
Validator-code typos are caught in tests; runtime crash isn't worth it over the existing error log + retry.

PR #195, opened 2026-04-24
swap_commands: use plain click in quote/resume

LandynDev:
## Summary
- `quote.py` and `resume.py` did `import rich_click as click`, but `rich_click` isn't in `pyproject.toml` — `alw --help` crashes with `ModuleNotFoundError: No module named 'rich_click'` on a clean install.
- Every other CLI file imports plain `click`; neither of these two uses anything `rich_click`-specific (decorators + `click.prompt` only), so switching to `click` matches the rest of the codebase with no behavior change.

## Test plan
- [ ] `uv sync && source .venv/bin/activate`
- [ ] `alw --help` runs cleanly
- [ ] `alw swap quote --help` / `alw swap resume-reservation --help` render
- [ ] `ruff check allways/` passes

### DIFF
--- a/allways/cli/swap_commands/quote.py
+++ b/allways/cli/swap_commands/quote.py
@@ -2,7 +2,7 @@
 
 from decimal import Decimal
 
-import rich_click as click
+import click
 from rich.table import Table
 
 from allways.chains import SUPPORTED_CHAINS, canonical_pair, get_chain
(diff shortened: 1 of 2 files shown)


dataCenter430:
Thanks for addressing this issue. @LandynDev 👍 
The problem was coming from the rich_click module, since it was unpacked in package.json.

PR #207, opened 2026-04-25
fix(btc): re-anchor against canonical chain at verify time

### LINKED ISSUE #182: No BTC reorg detection — single-point-in-time confirmation check
`verify_miner_fulfillment` reads `confirmations` once and then votes — both in the RPC path ([allways/chain_providers/bitcoin.py:169-177](allways/chain_providers/bitcoin.py#L169-L177)) and the Blockstream path ([allways/chain_providers/bitcoin.py:236-256](allways/chain_providers/bitcoin.py#L236-L256)).

Between that read and vote inclusion (12s+ cadence) a reorg can drop the tx. Bitcoin Core's `confirmations` field *does* update on reorgs, but the validator has already voted by then. Blockstream's `status.confirmed` is self-reported by whichever instance answered — no canonical-chain guarantee.

**Direction:** fetch the blockhash on every check and require that block to still be on the canonical chain at vote time; require min-confirmations to persist across at least two polls before voting.

dripsmvcp:
## Summary

Closes #182. `verify_miner_fulfillment` read BTC confirmations once and voted, in both the Core RPC path and the Blockstream path. Between the read and vote inclusion (12s+), a reorg can drop the tx — Core's `confirmations` updates on reorg but the validator has already voted; Blockstream's `status.confirmed` is self-reported by whichever instance answered.

After computing confirmations, fetch the canonical blockhash at the tx's height (`getblockhash` on RPC, `GET /block-height/<h>` on Blockstream) and compare to the tx's recorded `blockhash`. Mismatch → tx's block isn't on the canonical chain → drop to `confirmed=False` so the validator doesn't vote.

## Test plan

- [ ] Existing BTC verification tests pass
- [ ] Unit test: canonical-match path keeps `confirmed=True`
- [ ] Unit test: blockhash mismatch flips to `confirmed=False` and emits warning
- [ ] Manual: simulate reorg in regtest, verify validator abstains


dripsmvcp:

anderdc:
Fix conflicts.

PR #208, opened 2026-04-25
fix: reject NaN/inf rates in commitment parser

### LINKED ISSUE #184: NaN / inf rates bypass commitment parser
In [allways/commitments.py:40-42](allways/commitments.py#L40-L42), `float("inf")` and `float("nan")` parse without raising, and the surrounding `except` only catches `ValueError` / `IndexError` — so neither value is rejected.

**Impact:** an `inf` rate sorts to the top of crown-holder ranking; a `NaN` rate silently drops out of comparisons (all comparisons with NaN are false). Easy to exploit, easy to fix.

**Direction:** after `float(...)`, assert `math.isfinite(rate) and rate > 0`.

dripsmvcp:
## Summary

Closes #184. `float('nan')` and `float('inf')` don't raise, and the surrounding `except` only catches `ValueError` / `IndexError`, so neither was rejected by the commitment parser. An `inf` rate sorted to the top of crown-holder ranking; a `NaN` rate dropped silently out of comparisons (`NaN == NaN` is False).

After parsing both `rate` and `counter_rate`, reject any value that isn't `math.isfinite` and `> 0`.

## Test plan

- [ ] Existing commitment parser tests pass
- [ ] Unit test: `'inf'` / `'-inf'` / `'nan'` rate strings return `None`
- [ ] Unit test: `0` and negative rates still rejected
- [ ] Unit test: legitimate finite positive rates still parse

### DIFF
--- a/allways/commitments.py
+++ b/allways/commitments.py
@@ -1,5 +1,6 @@
 """Shared commitment parsing logic — used by validator, miner, and CLI."""
 
+import math
 from typing import List, Optional
 
 import bittensor as bt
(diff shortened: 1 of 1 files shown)


dripsmvcp:

dripsmvcp:

PR #209, opened 2026-04-25
fix: clear request on quorum failure, not just success

### LINKED ISSUE #185: Vote recorded before `on_quorum` runs — failed quorum strands the voter
In [smart-contracts/ink/lib.rs:230-233](smart-contracts/ink/lib.rs#L230-L233), `record_vote` writes at `:230`, then `on_quorum` runs at `:232`, and `clear_request` only runs at `:233` on success.

If the closure returns `Err`, the vote is persisted on a request that won't be cleared until expiry, and the validator can't re-vote in the meantime (`AlreadyVoted`).

Low-frequency in practice — `on_quorum` closures mostly re-check state that was already checked — but real, and the window grows with any future closure that can fail on post-quorum invariants.

**Direction:** either run `on_quorum` before recording the vote, or clear the request on error as well as success.

dripsmvcp:
## Summary

Closes #185. `consensus_vote` and `consensus_swap_vote` ran `on_quorum` first and only called `clear_request` on success. If the closure returned `Err` (e.g. a post-quorum invariant fails), the vote stayed persisted on a request that wouldn't be cleared until expiry, and the validator who cast the quorum-reaching vote could no longer re-vote (`AlreadyVoted`).

Run `on_quorum` and clear unconditionally; propagate the error after cleanup so storage stays consistent regardless of the closure's outcome.

## Test plan

- [ ] Existing consensus tests pass
- [ ] Unit test: `on_quorum` returning `Err` still clears `request_voters` / `request_hash` / `pending_swap_votes`
- [ ] Unit test: validator can re-vote on a fresh request after a failed quorum closure
- [ ] Unit test: original error from `on_quorum` is still surfaced to the caller

### DIFF
--- a/smart-contracts/ink/lib.rs
+++ b/smart-contracts/ink/lib.rs
@@ -229,8 +229,13 @@ mod allways_swap_manager {
             };
             let votes = self.record_vote(id, caller)?;
             if votes >= self.get_required_votes() {
-                on_quorum(self)?;
+                // Clear the request whether on_quorum succeeds or fails.
+                // Without this, a closure that returns Err leaves the vote
+                // persisted on a request that won't be cleared until expiry,
+                // blocking the validator from re-voting (AlreadyVoted).
+                let result = on_quorum(self);
                 self.clear_request(miner, req_type);
+                result?;
             }
             Ok(())
         }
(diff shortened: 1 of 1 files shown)


dripsmvcp:

PR #210, opened 2026-04-25
fix: higher confirmation threshold for destination leg

### LINKED ISSUE #187: Same confirmation threshold for user-source and miner-destination txs
[allways/chains.py:28](allways/chains.py#L28) hardcodes `BTC.min_confirmations = 3` globally, applied to both legs of a swap.

The two directions have asymmetric reorg risk:
- **User → miner** deposits: the miner bears the reorg risk.
- **Miner → user** fulfillment: the user bears the reorg risk.

They should not share a threshold. Bitcoin's de-facto standard for high-value receive is 6 confirmations; 3 is defensible for the leg where you're accepting inbound funds from a counterparty you already have collateral against, but weak for the leg where the user has already given up their funds.

**Direction:** split into `source_min_confirmations` (can stay at 3) and `destination_min_confirmations` (raise to 6, or make it scale with swap size).

dripsmvcp:
## Summary

Closes #187. `chains.py` hardcoded a single `min_confirmations` for both legs of a swap, but the two directions have asymmetric reorg risk:

- **User → miner** (source leg): miner bears reorg risk → 3 confs is fine
- **Miner → user** (destination leg): user has already given up source funds → reorg falls on them, 6 confs is the de-facto BTC standard

Adds `destination_min_confirmations` to `ChainDefinition` (defaults to 0, meaning "use `min_confirmations`"); set to 6 for BTC. `SwapVerifier.verify_tx` now takes `is_destination` and applies the higher threshold on the miner→user leg via `get_destination_min_confirmations(chain)`. Source-leg verification, CLI, and axon UX still use the original `min_confirmations`, so behavior is unchanged for those paths.

## Test plan

- [ ] Existing chain-verification tests pass
- [ ] Unit test: source leg accepts at 3 BTC confirmations (unchanged)
- [ ] Unit test: destination leg rejects at 3 confirmations, accepts at 6
- [ ] Unit test: TAO swaps unaffected (no `destination_min_confirmations` set → falls back to `min_confirmations`)
- [ ] Verify `verify_miner_fulfillment` passes `is_destination=True` only on the dest leg

### DIFF
--- a/allways/chains.py
+++ b/allways/chains.py
@@ -14,7 +14,8 @@ class ChainDefinition:
     decimals: int  # Precision (e.g. 8 for BTC, 9 for TAO)
     env_prefix: str  # .env variable prefix (e.g. 'BTC' -> BTC_RPC_URL)
     seconds_per_block: int = 12  # Average block time on this chain
-    min_confirmations: int = 1  # Minimum confirmations before accepting a transaction
+    min_confirmations: int = 1  # Default — source leg (user→miner): miner bears reorg risk
+    destination_min_confirmations: int = 0  # 0 = use min_confirmations; user→miner direction
 
 
 # ─── Supported Chains ────────────────────────────────────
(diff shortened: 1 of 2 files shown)


dripsmvcp:

anderdc:
#187 was already resolved in main (closed completed) — the destination-leg threshold is already in place. Closing.

PR #212, opened 2026-04-25
fix: retain failed blocks for retry instead of advancing

### LINKED ISSUE #201: [S1] `event_watcher` silently drops events on transient RPC failure
**Location:** [allways/validator/event_watcher.py:313-322](allways/validator/event_watcher.py#L313-L322), [:324-332](allways/validator/event_watcher.py#L324-L332)

### Bug
`process_block` catches all exceptions and returns at DEBUG log level. `sync_to` then unconditionally advances `self.cursor = end`. Any block that failed to fetch during a sync window is never reprocessed.

### Impact
- Missed `SwapCompleted` / `SwapTimedOut` → miners permanently flagged busy in scoring.
- Missed `MinerActivated` → active-set drift.

Self-healing via bootstrap on restart, but in-process the scoring replay is wrong until then.

### Fix
On per-block exception either raise to caller, or keep a pending-retry list and only advance cursor past blocks that succeeded.

dripsmvcp:
## Summary

Closes #201. `process_block` caught all exceptions at DEBUG, and `sync_to` unconditionally advanced `cursor = end` past them. Any block that failed to fetch during a sync window was never reprocessed in-process — missed `SwapCompleted`/`SwapTimedOut` flagged miners busy until restart-time bootstrap healed it; missed `MinerActivated` caused active-set drift over the same window.

`process_block` now returns `True` / `False`; `sync_to` advances cursor only as far as the last cleanly processed block, so failed blocks are retried on the next forward step.

## Test plan

- [ ] Existing event-watcher tests pass
- [ ] Unit test: clean run advances cursor to `end` (unchanged)
- [ ] Unit test: mid-window RPC failure stops cursor at last-good block
- [ ] Unit test: subsequent `sync_to` retries the failed block once it recovers
- [ ] Verify `prune_old_events` still runs even on partial sync

### DIFF
--- a/allways/validator/event_watcher.py
+++ b/allways/validator/event_watcher.py
@@ -312,24 +312,29 @@ class ContractEventWatcher:
 
     def sync_to(self, current_block: int) -> None:
         """Catch up from cursor to ``current_block`` in MAX_BLOCKS_PER_SYNC
-        chunks so a long outage doesn't freeze the forward loop."""
+        chunks so a long outage doesn't freeze the forward loop. Only advance
+        cursor past blocks that processed cleanly — a transient RPC failure
+        on one block must not strand its events."""
         if current_block <= self.cursor:
             return
         end = min(current_block, self.cursor + MAX_BLOCKS_PER_SYNC)
+        last_ok = self.cursor
         for block_num in range(self.cursor + 1, end + 1):
-            self.process_block(block_num)
-        self.cursor = end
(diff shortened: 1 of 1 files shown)


dripsmvcp:

LandynDev:
Please rebase this asap @dripsmvcp

LandynDev:
This work was addressed in https://github.com/entrius/allways/issues/240

PR #213, opened 2026-04-25
fix(event-watcher): don't double-count in-flight swaps after restart

### LINKED ISSUE #202: [S1] `event_watcher` double-counts in-flight swaps across restart
**Severity:** S1 (silent emission loss for affected miners)

**Location:** [allways/validator/event_watcher.py:285-311](allways/validator/event_watcher.py#L285-L311)

### Bug
Bootstrap seeds `+1 BusyEvent` at each in-flight swap's `initiated_block` (L296-297), then rewinds cursor to `current_block - SCORING_WINDOW_BLOCKS` (L304). `sync_to` replays blocks from `cursor+1` onward — including the *same* `SwapInitiated` event whose `+1` was just seeded — so `apply_busy_delta(+1)` fires a second time.

When the swap eventually resolves, the matching `-1` lands once, not twice, so the miner stays stuck at `open_swap_count = 1` forever. Permanently excluded from crown-time credit as "busy."

`SCORING_WINDOW_BLOCKS = 1200` (~4h); fulfillment timeouts are much shorter, so essentially every restart with in-flight swaps triggers this.

### Fix options
- (a) Seed bootstrap +1 at `cursor` rather than `init_block`, and skip replay events for swaps that were in the bootstrap set.
- (b) Bootstrap after `sync_to`, diff expected vs observed counts.
- (c) Set cursor to `current_block` (skip backfill) and accept loss of pre-restart crown credit.

dripsmvcp:
## Summary

Closes #202. Bootstrap seeded `+1 BusyEvent` for each in-flight swap at its `initiated_block`, then rewound cursor to `current_block - SCORING_WINDOW_BLOCKS`. `sync_to` then replayed the same `SwapInitiated` event, applying a second `+1`. The matching `-1` from terminal events fires only once, so the miner stayed stuck at `open_swap_count=1` forever — permanently excluded from crown-time credit.

Track the swap IDs seeded during bootstrap in `_bootstrap_initiated_ids`; in `apply_event`, skip the `SwapInitiated` replay for any ID in that set (and discard it from the set so a future legitimate event for the same ID would still count, though IDs are unique so this is defensive).

## Test plan

- [ ] Existing event-watcher tests pass
- [ ] Unit test: restart with in-flight swap → `open_swap_count` is 1 after bootstrap + sync (not 2)
- [ ] Unit test: terminal event resolves the count to 0
- [ ] Unit test: swap initiated *after* bootstrap still increments normally
- [ ] Unit test: a swap not in bootstrap set still increments via replay

### DIFF
--- a/allways/validator/event_watcher.py
+++ b/allways/validator/event_watcher.py
@@ -215,6 +215,9 @@ class ContractEventWatcher:
         self.active_miners: Set[str] = set()
         self.open_swap_count: Dict[str, int] = {}
         self.busy_events: List[BusyEvent] = []
+        # Swap IDs already counted as +1 by bootstrap. SwapInitiated events
+        # for these IDs replayed by sync_to are skipped so we don't double-count.
+        self._bootstrap_initiated_ids: Set[int] = set()
         self.active_events: List[ActiveEvent] = []
         self.active_events_by_hotkey: Dict[str, List[ActiveEvent]] = {}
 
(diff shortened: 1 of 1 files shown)


dripsmvcp:

LandynDev (approved):

LandynDev:
Taking #191 instead — same author, same fix, has tests, cleaner terminal-event cleanup.

PR #214, opened 2026-04-25
fix(swap): give user a path to surface an already-sent tx before retrying

### LINKED ISSUE #204: `sendtoaddress` retry could re-select UTXOs after first broadcast propagated
**Severity:** S3 (fragile boundary, not a silent double-send today)

**Location:** [allways/cli/swap_commands/swap.py:434-449](allways/cli/swap_commands/swap.py#L434-L449), [allways/chain_providers/bitcoin.py:605-634](allways/chain_providers/bitcoin.py#L605-L634)

Node-mode BTC send uses `sendtoaddress` (or `rpc_send_from_address`), returns `None` on any RPC exception. The outer retry at [swap.py:452-456](allways/cli/swap_commands/swap.py#L452-L456) prompts the user and retries. If the first broadcast actually landed but the RPC response was lost, Core's wallet locks the UTXOs — so retry's `listunspent` won't re-select them — and the retry either fails ("insufficient") or picks a different UTXO set producing a different tx. Lightweight mode with Blockstream has no such wallet-level UTXO locking.

**Fix:** After any send exception, query mempool/chain for a tx matching `(sender, recipient, amount)` before retrying.

---

dripsmvcp:
If the first sendtoaddress broadcast actually propagated but the RPC response was lost, Core's wallet locks the UTXOs so a retry either fails with insufficient funds or picks a different UTXO set producing a second tx. Lightweight mode has no wallet-level UTXO locking and is even more exposed.

Before retrying, warn the user about the locked-UTXO / second-send risk and prompt for an already-sent tx hash from their wallet/mempool. If they paste one that verifies, use it; otherwise fall through to retry. Keeps the existing flow for the common case (real RPC failure).

Closes #204

### DIFF
--- a/allways/cli/swap_commands/swap.py
+++ b/allways/cli/swap_commands/swap.py
@@ -451,6 +451,27 @@ def send_btc(chain_providers, config, to_address: str, amount_sat: int, from_add
         # Retry or fall through to manual
         if attempt < max_retries:
             console.print('[yellow]BTC send failed.[/yellow]')
+            console.print(
+                '[dim]If the first broadcast actually landed but the RPC response was lost,[/dim]\n'
+                '[dim]retrying may either fail with insufficient funds (UTXOs locked) or pick a[/dim]\n'
+                '[dim]different UTXO set and produce a second send. Check your wallet before retrying.[/dim]'
+            )
+            existing = click.prompt(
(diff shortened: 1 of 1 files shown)


dripsmvcp:

LandynDev:
Closed with reason in https://github.com/entrius/allways/issues/204

PR #231, opened 2026-04-29
fix(validator): only swallow recoverable transport errors in forward gather

### LINKED ISSUE #178: `asyncio.gather` swallows programming errors

jeffrey701:
```
Closes #178.

`confirm_miner_fulfillments` collected verifier results with
`asyncio.gather(..., return_exceptions=True)` and then matched
`isinstance(result, ProviderUnreachableError)` for the deferral case
and a catch-all `isinstance(result, Exception)` for everything else.
The catch-all reduced *every* exception type — including programming
errors like `AttributeError`, `KeyError`, `TypeError` and any future
typo in the verifier — to a logged "verification error" while the
forward loop kept running with silently-missing votes. Real bugs were
masked as flaky network conditions for the lifetime of the deployment.

## Fix

- Added `_verify_swallowing_only_recoverable(verifier, swap)`: an async
  wrapper that catches the explicit allow-list
  `ProviderUnreachableError` + `asyncio.TimeoutError` and converts them
  into return values, but lets every other exception propagate.
- `confirm_miner_fulfillments` now wraps each verifier call in this
  helper and uses plain `gather(...)` (no `return_exceptions=True`),
  so programming errors fail loud at the forward step boundary
  instead of being absorbed.
- Added explicit `asyncio.TimeoutError` handling in the result loop —
  previously this case was swept into the catch-all warning log; now
  it routes to the same "uncertain, defer" bucket as
  `ProviderUnreachableError`.
- Exposed `RECOVERABLE_VERIFY_EXCEPTIONS` and the helper at module
  level so the test suite can pin both the allow-list and the wrapper
  semantics directly.

## Tests

`tests/test_forward_verify_recoverable.py` adds 10 cases covering:
- the allow-list contents (`ProviderUnreachableError`,
  `asyncio.TimeoutError` are recoverable; `Exception`,
  `AttributeError`, `KeyError`, `TypeError`, `ValueError` are not);
- the wrapper passes True / False results through unchanged;
- recoverable exceptions are returned as values;
- programming errors (`AttributeError`, `KeyError`, `TypeError`)
  propagate unmasked.

`uv run pytest tests/` — 342 passed (10 new).
`uv run ruff check ...` — clean.

Closes #178
```

### DIFF
--- a/allways/validator/forward.py
+++ b/allways/validator/forward.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import asyncio
-from typing import TYPE_CHECKING, Set
+from typing import TYPE_CHECKING, Set, Tuple, Type, Union
 
 import bittensor as bt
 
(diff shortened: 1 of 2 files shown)


LandynDev:
Taking #192 instead — same author already filed surgical 6-line fix with same allow-list semantics.

PR #241, opened 2026-04-30
CLI: sanitize ALW_DENDRITE_TIMEOUT (reject nan/inf/zero/negative)

### LINKED ISSUE #240: CLI: sanitize ALW_DENDRITE_TIMEOUT env (reject nan/zero/negative)
### Description

**Severity:** S3

**Location:** `allways/cli/dendrite_lite.py` — `resolve_dendrite_timeout()`

### Steps to reproduce

```bash
uv run python -c "import os; os.environ['ALW_DENDRITE_TIMEOUT']='nan'; from allways.cli.dendrite_lite import resolve_dendrite_timeout; print(repr(resolve_dendrite_timeout(30)))"
```

### Expected behaviour

- Only finite timeouts `> 0`; else default.

### Actual behaviour

- Returns `nan`, `-1.0`, or `0.0` unchanged.

### Screenshots

Image

eureka0928:
## Summary

`resolve_dendrite_timeout` ([dendrite_lite.py:95-105](https://github.com/entrius/allways/blob/test/allways/cli/dendrite_lite.py#L95-L105)) returned `float(override)` directly, so `'nan'`, `'inf'`, `'0'`, and negative values slipped through to the dendrite call. None of those are useful overrides and each breaks differently:

- `nan` — `asyncio.wait_for` rejects `nan` deep inside the coroutine, surfacing as confusing "Operation cancelled" or hangs.
- `inf` / `-inf` — defeats the override's purpose; the CLI sits indefinitely on a dead validator.
- `0` / negative — every dendrite send times out instantly.

Add a `math.isfinite() and value > 0` guard, fall back to the caller's default, and emit `bt.logging.warning` so the operator knows their override was ignored. The existing `ValueError` fallback was silent — same warning treatment now, so a typo'd `ALW_DENDRITE_TIMEOUT=Inifity` is debuggable instead of mysteriously inert.

### Reproducer (from the issue)
```bash
$ uv run python -c "import os; os.environ['ALW_DENDRITE_TIMEOUT']='nan'; \
    from allways.cli.dendrite_lite import resolve_dendrite_timeout; \
    print(repr(resolve_dendrite_timeout(30)))"

# Before:  nan
# After:   30.0   (and a warning logged about the rejected value)
```

## Test plan

- [x] `pytest tests/test_dendrite_lite.py` — 12 tests covering: unset/empty env → default; integer / fractional / whitespace-padded values pass through; six bad-value cases (`nan`, `inf`, `-inf`, `0`, `-1`, `foo`) fall back; the *caller's* `default` is honored on fallback (not a hardcoded 30)
- [ ] `pytest tests/` — full suite green
- [ ] `ruff check` + `ruff format` clean
- [ ] Smoke: `ALW_DENDRITE_TIMEOUT=nan alw status` runs normally; warning in logs identifies the rejected value
- [ ] Smoke: `ALW_DENDRITE_TIMEOUT=60 alw swap quote ...` uses the 60s override (no warning)
- [ ] Smoke: `ALW_DENDRITE_TIMEOUT=0 alw status` does **not** instantly time out

Fixes #240.

### DIFF
--- a/allways/cli/dendrite_lite.py
+++ b/allways/cli/dendrite_lite.py
@@ -93,13 +93,28 @@ def broadcast_synapse(
 
 
 def resolve_dendrite_timeout(default: float) -> float:
-    """Honor ALW_DENDRITE_TIMEOUT as an override for slow chains (e.g. testnet)."""
+    """Honor ALW_DENDRITE_TIMEOUT as an override for slow chains (e.g. testnet).
+
+    Rejects ``nan``, ``inf``, and non-positive values so a typo in the env
+    var doesn't silently break dendrite calls (issue #240). ``float()`` alone
+    accepts all three, and the dendrite call would either hang on ``nan``,
+    bypass the override on ``inf``, or fail every send instantly on ``0``/
+    negative — none of which are useful, so we fall back to ``default``.
+    """
+    import math
     import os
 
     override = os.environ.get('ALW_DENDRITE_TIMEOUT')
     if not override:
         return default
     try:
-        return float(override)
+        value = float(override)
     except ValueError:
+        bt.logging.warning(f'ALW_DENDRITE_TIMEOUT={override!r} is not a number, using default {default}')
         return default
+    if not math.isfinite(value) or value <= 0:
+        bt.logging.warning(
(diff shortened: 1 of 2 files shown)


LandynDev:
This work was addressed in https://github.com/entrius/allways/issues/240

PR #245, opened 2026-04-30
CLI: write config atomically; warn loudly on unreadable config

### LINKED ISSUE #244: CLI: alw config set writes config non-atomically; partial write silently reverts to defaults
### Description

**Severity:** S2 (silent data loss — user's network/contract reverts to mainnet defaults)

**Location:** [`allways/cli/main.py:140-159`](https://github.com/entrius/allways/blob/main/allways/cli/main.py#L140-L159)

```python
ALLWAYS_DIR.mkdir(parents=True, exist_ok=True)
config = {}
if CONFIG_FILE.exists():
    try:
        config = json.loads(CONFIG_FILE.read_text())
    except json.JSONDecodeError:
        ...
...
CONFIG_FILE.write_text(json.dumps(config, indent=2))
```

`write_text` is a truncate-then-write — any interrupt mid-write (Ctrl+C, OOM, power loss, container kill) leaves a truncated/empty JSON file. The reader at [`load_cli_config`](https://github.com/entrius/allways/blob/main/allways/cli/swap_commands/helpers.py#L163-L170) then swallows `JSONDecodeError` and returns `{}`:

```python
def load_cli_config() -> dict:
    if not CONFIG_FILE.exists():
        return {}
    try:
        return json.loads(CONFIG_FILE.read_text())
    except Exception:
        return {}
```

Result: every config key (`network`, `wallet`, `hotkey`, `contract-address`, `netuid`) silently reverts to its default. A user who had `network=local` configured for dev will, on the next command, hit **finney mainnet with the mainnet contract address** and have no warning.

The same package already has the atomic-write pattern — `save_pending_swap` uses `tempfile.mkstemp` + `os.replace` ([helpers.py:279-291](https://github.com/entrius/allways/blob/main/allways/cli/swap_commands/helpers.py#L279-L291)).

### Steps to reproduce

1. `alw config set network local`
2. Manually corrupt the file to simulate an interrupted write: `printf '{"network": "lo' > ~/.allways/config.json`
3. `alw config` → renders empty (no warning).
4. Any command (`alw view rates`, `alw swap now`) now points at **finney**, not the configured `local`.

### Expected behaviour

- `alw config set` writes via tempfile + `os.replace` so a partial write cannot corrupt the file.
- `load_cli_config` warns loudly on `JSONDecodeError` rather than silently returning `{}`.

### Actual behaviour

- Truncated config silently treated as empty; user runs against mainnet without notice.

### Fix

- Mirror the `save_pending_swap` pattern in `config_set`:

  ```python
  fd, tmp_path = tempfile.mkstemp(dir=ALLWAYS_DIR, suffix='.tmp')
  try:
      with os.fdopen(fd, 'w') as f:
          f.write(json.dumps(config, indent=2))
      os.replace(tmp_path, CONFIG_FILE)
  except Exception:
      if os.path.exists(tmp_path):
          os.unlink(tmp_path)
      raise
  ```

- Optionally: surface a `[yellow]Warning: existing config was unreadable[/yellow]` from `load_cli_config` instead of returning `{}`.

eureka0928:
## Summary

`config_set` ([main.py:140-159](https://github.com/entrius/allways/blob/test/allways/cli/main.py#L140-L159)) used to truncate-then-write `~/.allways/config.json` in place via `write_text`. A mid-write interrupt (Ctrl+C, OOM, container kill, power loss) leaves a truncated JSON file. `load_cli_config` ([helpers.py:163-170](https://github.com/entrius/allways/blob/test/allways/cli/swap_commands/helpers.py#L163-L170)) then silently swallowed the parse error and returned `{}`, reverting every key (`network`, `contract-address`, `wallet`, `hotkey`, `netuid`) to mainnet defaults — a dev who'd configured `network=local` would silently end up on **finney with the mainnet contract address**, no warning.

Two changes:

- **`config_set` now writes atomically** via `tempfile.mkstemp` + `os.replace`, mirroring [`save_pending_swap`](https://github.com/entrius/allways/blob/test/allways/cli/swap_commands/helpers.py#L279-L291) (already in the codebase). Either the new file lands fully or the old file is untouched. On any write failure the temp file is unlinked.
- **`load_cli_config` now prints a Rich warning** when the file exists but won't parse, instead of silently swallowing the exception. Callers still get `{}` so nothing crashes — the user just knows the config didn't load and what to do (`alw config set ...` to repair).

### Reproducer (from the issue)
```
$ alw config set network local           # state was good
$ printf '{"network": "lo' > ~/.allways/config.json   # simulate interrupted write
$ alw view rates
# Before: silently runs against finney mainnet
# After:  Warning: ~/.allways/config.json is unreadable (...); using defaults.
#         Run `alw config set ...` to repair.
```

## Test plan

- [x] `pytest tests/test_config_atomic_write.py` — 8 tests covering: happy-path writes a complete file; no `.tmp` leftover on success; existing config preserved when `os.replace` is interrupted (the S2 reproducer fixed); temp file cleaned up on write failure; missing file returns `{}` silently; valid file parses correctly; truncated file warns and returns `{}`; empty file warns and returns `{}`
- [ ] `pytest tests/` — full suite green
- [ ] `ruff check` + `ruff format` clean
- [ ] Smoke: `alw config set network local`, then `printf '{}' > ~/.allways/config.json` corrupted state — next command warns instead of silently using mainnet
- [ ] Smoke: `Ctrl+C` during a real `alw config set` leaves the previous config intact

Fixes #244.

### DIFF
--- a/allways/cli/main.py
+++ b/allways/cli/main.py
@@ -15,6 +15,7 @@ import os  # noqa: E402
 # Prevent bittensor from hijacking --help via its argparse config.
 # Must happen before any bittensor import.
 import sys as _sys
+import tempfile  # noqa: E402
 
 _saved_argv = _sys.argv[:]
 _sys.argv = [_sys.argv[0]]
@@ -156,7 +157,20 @@ def config_set(key: str, value: str):
     old_value = config.get(key)
     config[key] = value
 
-    CONFIG_FILE.write_text(json.dumps(config, indent=2))
+    # Atomic write — Ctrl+C, OOM, or container kill mid-write used to
+    # truncate the file, which load_cli_config then silently treated as
+    # empty, reverting every key (network, contract-address, etc.) to
+    # mainnet defaults (issue #244). tempfile + os.replace either lands
+    # the new file fully or leaves the old one untouched.
+    fd, tmp_path = tempfile.mkstemp(dir=ALLWAYS_DIR, suffix='.tmp')
+    try:
+        with os.fdopen(fd, 'w') as f:
+            f.write(json.dumps(config, indent=2))
+        os.replace(tmp_path, CONFIG_FILE)
+    except Exception:
+        if os.path.exists(tmp_path):
+            os.unlink(tmp_path)
+        raise
 
     display = value
     if key == 'network' and value in KNOWN_NETWORKS:
--- a/allways/cli/swap_commands/helpers.py
+++ b/allways/cli/swap_commands/helpers.py
@@ -161,12 +161,23 @@ def from_rao(amount_rao: int) -> float:
 
 
 def load_cli_config() -> dict:
-    """Load CLI configuration from ~/.allways/config.json."""
+    """Load CLI configuration from ~/.allways/config.json.
+
+    Returns ``{}`` on a missing or unreadable file so callers don't crash,
+    but surfaces a Rich warning when the file exists and fails to parse —
+    silently swallowing corruption used to hide a partial-write recovery
+    that reverted ``network``/``contract-address`` to mainnet defaults
+    (issue #244).
+    """
     if not CONFIG_FILE.exists():
         return {}
     try:
         return json.loads(CONFIG_FILE.read_text())
-    except Exception:
+    except Exception as e:
+        console.print(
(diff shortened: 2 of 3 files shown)


LandynDev:
Left comments on this work in the tagged issue #244

PR #251, opened 2026-05-01
collateral deposit: shrink gas buffer to 0.02 TAO

LandynDev:
## Summary
- Drops `MIN_BALANCE_FOR_TX_RAO` from 0.25 TAO → 0.02 TAO. Real extrinsic fees are sub-millitao; the old buffer was rejecting deposits that would have comfortably gone through.
- Reworded the insufficient-balance error in `alw collateral deposit` so it explains we pre-check the gas buffer to avoid burning fees on a reverted tx.

## Test plan
- [ ] `alw collateral deposit --amount X` where `free = X + 0.02 TAO` succeeds (previously would have required `X + 0.25 TAO`).
- [ ] Error path: `free < amount + 0.02 TAO` prints the new wording.


LandynDev:
Reopening on a clean branch off test — see new PR.

PR #253, opened 2026-05-01
fix: keep pending_confirms row alive when finalize lands same step as purge

anderdc:
## Summary

Fixes the testnet swap that stranded 0.2 TAO at `5DFTjn5q…WazE` (reservation expired at block 7024162; validator's next forward step at 7024166 silently purged the row before finalize could run).

**Root cause** (`allways/validator/forward.py`): the forward loop called `purge_expired_pending_confirms()` *before* `initialize_pending_user_reservations()`. `maybe_finalize_reservation` is only reachable from inside the per-row loop in `initialize_pending_user_reservations`, so any propose whose first finalize-eligible step landed at-or-after the original `reserved_until` was lost: the row was deleted before finalize ever got a chance.

## Changes

- **Reorder**: `initialize_pending_user_reservations` now runs before `purge_expired_pending_confirms`. Finalize gets a chance on rows whose original deadline has just lapsed.
- **`maybe_finalize_reservation` now returns `Optional[int]`** (the applied `target_block`) instead of `bool`. `try_extend_reservation` uses the returned target to call `state_store.update_reserved_until` in-line, so the same-step purge sees the bumped deadline. Without this, the race re-emerges after the reorder because `ReservationExtensionFinalized` isn't picked up until next step's event sync.
- **Defensive floor on tier-0 propose targets**: `target_block >= current_block + CHALLENGE_WINDOW_BLOCKS + VALIDATOR_FORWARD_STEP_BLOCKS_ESTIMATE`, rounded up to an `EXTENSION_BUCKET_BLOCKS` boundary. With the current chain configs (TAO seconds_per_block=12, BTC=600) the natural bucketed target already clears this floor, so this is a forward-looking guard for tighter padding or fast-block chains.

## Caveat (worth surfacing for review)

The contract rejects `finalize_extend_reservation` when `reservation.reserved_until < current` (`smart-contracts/ink/lib.rs:688-692`). This patch ensures the row survives whenever finalize *can* succeed in-step, but it does not widen the window when forward cadence > `reserved_until - propose_block - CHALLENGE_WINDOW_BLOCKS`. On the failing testnet swap (forward step ~19 blocks, finalize window ~7 blocks) finalize would still be rejected by the contract. Widening that window requires one of:

- bumping `reservation_ttl` on the contract (admin call, no code change)
- raising `EXTEND_THRESHOLD_BLOCKS` so propose lands earlier in the reservation
- speeding up validator forward cadence

Out of scope here — happy to follow up in a separate PR/operational ticket.

## Test plan

- [x] `pytest tests/test_optimistic_extensions.py tests/test_pending_confirm_queue.py tests/test_validator_rejections.py tests/test_event_watcher.py tests/test_chains.py tests/test_swap_tracker.py` — 100 passed
- [x] Updated `TestMaybeFinalizeReservation` to assert on returned target; added `test_returns_none_when_contract_call_fails`.
- [x] Existing `test_update_reserved_until_prevents_stale_purge` already covers the state_store half of the contract.
- [ ] Manual: replay a near-deadline propose+finalize against testnet on a single-validator subnet, verify the row survives across the purge.

### DIFF
--- a/allways/constants.py
+++ b/allways/constants.py
@@ -75,6 +75,11 @@ MAX_EXTENSION_BLOCKS = 250  # client-side cap, mirrors the contract's hard cap
 # smart-contracts/ink/lib.rs. Validators gate finalize calls on this locally
 # to avoid known-doomed txs; the contract is authoritative.
 CHALLENGE_WINDOW_BLOCKS = 8
+# Conservative upper bound on subtensor blocks elapsed per validator forward
+# step (base poll + per-step work + jitter). Used as a safety margin when
+# sizing extension targets so a single delayed step doesn't strand a propose
+# whose finalize window opens past the original reservation deadline.
+VALIDATOR_FORWARD_STEP_BLOCKS_ESTIMATE = 20
 
 # Tiered escalation. First extension fires on tx visibility alone (mempool
 # OK) and buys time for one block; second extension requires ≥1 confirmation
--- a/allways/validator/forward.py
+++ b/allways/validator/forward.py
@@ -41,18 +41,24 @@ async def forward(self: Validator) -> None:
 
     clear_provider_caches(self)
 
-    # Sync events before purge so any ReservationExtensionFinalized in the
-    # last block writes the new reserved_until back to state_store before the
-    # purge sweep runs — otherwise a row whose contract deadline just bumped
-    # would still be deleted at its stale (original) reserved_until.
+    # Sync events first so ReservationExtensionFinalized writes from the
+    # previous block reach state_store before the per-row finalize/init loop
+    # reads them.
     try:
         self.event_watcher.sync_to(self.block)
     except Exception as e:
         bt.logging.warning(f'Event watcher sync failed: {e}')
 
-    self.state_store.purge_expired_pending_confirms()
-
+    # Init runs *before* purge so maybe_finalize_reservation gets a chance to
+    # fire on rows whose original reserved_until has just lapsed. A successful
+    # finalize bumps state_store.reserved_until in-line (see
+    # try_extend_reservation), so the subsequent purge sees the fresh deadline
+    # and leaves the row alone. Without this ordering, any propose whose
+    # finalize-eligible step lands at-or-after the original reserved_until
+    # would be silently lost to the purge before init ever sees the row.
     initialize_pending_user_reservations(self)
+
+    self.state_store.purge_expired_pending_confirms()
 
     poll_commitments(self)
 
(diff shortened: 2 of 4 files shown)


LandynDev (commented):
Read end-to-end. The reorder + in-line `update_reserved_until` is the right shape for the bug we hit on the testnet swap that stranded the user's 0.2 TAO — without the in-line write, swapping the two calls just moves the race one block over. Everything below is small.

## What actually fixes the reported bug
- `forward.py` order swap so `initialize_pending_user_reservations` runs before `purge_expired_pending_confirms`.
- `maybe_finalize_reservation` returning `Optional[int]` and `try_extend_reservation` writing the new deadline to `state_store.update_reserved_until` in the same step.

That pair is sufficient. Tests cover both halves (existing `test_update_reserved_until_prevents_stale_purge` + the new `test_returns_none_when_contract_call_fails`).

## Things to address before merge

1. **Same race exists on the timeout side, not fixed here.** `extend_fulfilled_near_timeout` calls `maybe_finalize_timeout` (still returns `bool`) and `enforce_swap_timeouts` runs immediately after using the snapshot of `tracker.timeout_block` from before finalize. A swap whose finalize lands at-or-just-past its original `timeout_block` will get voted timed-out by the same forward step, even though the contract just bumped the deadline. Suggest mirroring the reservation fix: `maybe_finalize_timeout → Optional[int]`, `extend_fulfilled_near_timeout` calls `swap_tracker.update_timeout_block(swap.id, returned_target)` in-line, and `enforce_swap_timeouts` re-reads the fresh value. Can be a follow-up if you want to keep this PR scoped, but worth filing.

2. **Defensive floor only on `maybe_propose_reservation`, not `maybe_propose_timeout`.** Symmetry argument — same chain math, same fragile-on-fast-blocks risk. Drop it in both, or note the asymmetry in the constant's docstring.

3. **The defensive floor is a no-op on every supported chain today.** With `EXTENSION_BUCKET_BLOCKS=30` and `EXTENSION_PADDING_SECONDS=300`:
   - TAO source: tier-0 already produces 30 blocks; floor of 28 → bucket-rounded back to 30. No change.
   - BTC source: tier-0 = 90 blocks, tier-1 = 150 blocks. Floor of 28 is well under both. No change.

   Worth mentioning explicitly in the PR description so a future reader doesn't think the floor is what fixes the bug — the reorder + in-line write is. As written the `VALIDATOR_FORWARD_STEP_BLOCKS_ESTIMATE = 20` only kicks in for hypothetical sub-12s-block chains or if someone drops `EXTENSION_PADDING_SECONDS`. Fine to keep as a tripwire, just don't sell it as load-bearing.

## Caveat in the PR description is correct

> the contract still rejects finalize when `reservation.reserved_until < current` … on the failing testnet swap finalize would still be rejected.

Confirmed against `smart-contracts/ink/lib.rs:688-692`. The actual operational unblock for that incident is one of: bump `reservation_ttl` (admin call, no code), raise `EXTEND_THRESHOLD_BLOCKS` so propose lands earlier, or speed up validator forward cadence. None of those belong in this PR.

## Nits
- `min_safe_target` block in `maybe_propose_reservation` (lines 85-93) — the `(min_safe_target - current_block + EXTENSION_BUCKET_BLOCKS - 1) // EXTENSION_BUCKET_BLOCKS * EXTENSION_BUCKET_BLOCKS + current_block` could just call `compute_extension_target(from_chain_id, ...)` with a derived `remaining` so all the bucket-rounding lives in one place. Cosmetic.
- `VALIDATOR_FORWARD_STEP_BLOCKS_ESTIMATE` belongs next to `EXTEND_THRESHOLD_BLOCKS` in `constants.py`, not below `CHALLENGE_WINDOW_BLOCKS`. They're operationally coupled.

LGTM modulo (1) being filed and (3) being mentioned in the description.

LandynDev (approved):

PR #254, opened 2026-05-01
validator: mirror finalize-cache write for swap timeout side

LandynDev:
## Why

#253 closed the same-step finalize/purge race on the **reservation** side (`maybe_finalize_reservation` returns the applied target; `try_extend_reservation` writes it to `state_store.update_reserved_until` in-line so the upstream purge sees the bumped deadline).

The **timeout** side has the identical race against a different cache:

```
forward()
  ├─ extend_fulfilled_near_timeout(self)   # calls maybe_finalize_timeout
  └─ enforce_swap_timeouts(self, tracker, …)  # reads tracker.timeout_block
```

`maybe_finalize_timeout` returns `bool`, so even when the contract just bumped the timeout, `swap_tracker` doesn't learn until the *next* step's `event_watcher.sync_to` decodes `TimeoutExtensionFinalized`. `enforce_swap_timeouts` runs immediately after the loop, reads the pre-finalize `swap.timeout_block`, votes timeout on a swap whose contract-side deadline has already moved, and the contract rejects the vote. Wasted RPC + a noisy `Swap N: timed out` warning.

## Change

Single commit, mirrors the shape #253 used for reservations:

- `maybe_finalize_timeout(...) -> Optional[int]` (was `bool`) — returns the applied `target_block` on success.
- `extend_fulfilled_near_timeout` calls `tracker.update_timeout_block(swap.id, finalized_target)` immediately so `enforce_swap_timeouts` reads the bumped deadline rather than the stale one.

That's it. No reorder, no new constants, no defensive floor — those all live on the reservation side via #253.

## Test plan

- [x] `pytest tests/test_optimistic_extensions.py tests/test_pending_confirm_queue.py tests/test_validator_rejections.py tests/test_event_watcher.py tests/test_swap_tracker.py` — 85 passed
- [x] Added `TestMaybeFinalizeTimeout::test_returns_none_when_contract_call_fails` and updated the two existing finalize tests to assert on the returned target.
- [ ] Manual: testnet swap that crosses its `timeout_block` mid-step with a pending timeout extension — should finalize and keep the swap alive instead of voting timed-out.

## History

This branch originally carried the same reservation-side fix as #253 plus the timeout mirror. After #253 merged it was rebased onto `test` and the duplicated reservation commits were dropped — what's left is the one-commit timeout-side mirror.

### DIFF
--- a/allways/validator/forward.py
+++ b/allways/validator/forward.py
@@ -410,11 +410,16 @@ def extend_fulfilled_near_timeout(self: Validator) -> None:
         swap_label = f'{swap.from_chain.upper()}->{swap.to_chain.upper()}'
         ctx = f'Swap #{swap.id} [{swap_label}]'
 
-        self.optimistic_extensions.maybe_finalize_timeout(
+        finalized_target = self.optimistic_extensions.maybe_finalize_timeout(
             swap_id=swap.id,
             current_block=current_block,
             challenge_window_blocks=CHALLENGE_WINDOW_BLOCKS,
         )
+        if finalized_target is not None:
+            # Same-step write so enforce_swap_timeouts (which runs immediately
+            # after this loop) reads the bumped deadline rather than the
+            # pre-finalize value the next event sync would otherwise carry in.
+            tracker.update_timeout_block(swap.id, finalized_target)
 
         provider = self.chain_providers.get(swap.to_chain)
         if not provider or not swap.to_tx_hash:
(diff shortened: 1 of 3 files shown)


LandynDev:
Closing — kimbo's PR #253 is still in progress; deferring to it. Branch left in place in case the timeout-side mirror or the simplification angle is useful later.

LandynDev:
Reopening per request.

PR #304, opened 2026-05-10
Therefactor: extract _metagraph_hotkey_to_uid helper in commitments.py

### LINKED ISSUE #169: `read_miner_commitment` and `read_miner_commitments` independently build `hotkey_to_uid`
[allways/commitments.py:123](allways/commitments.py#L123) and [allways/commitments.py:144](allways/commitments.py#L144) both compute the same `{hotkey: uid}` map. Extract `_metagraph_hotkey_to_uid(metagraph)` once so callers cannot accidentally use a stale map.

Tet-9:
## Summary

`read_miner_commitment` and `read_miner_commitments` both independently
built the same `{hotkey: uid}` map inline:

- `commitments.py:138` — `{metagraph.hotkeys[u]: u for u in range(metagraph.n.item())}`
- `commitments.py:160` — identical expression with different loop variable

Duplicate construction means a future change to the map logic must be
applied twice, and callers could accidentally use a stale copy.

## Fix

Extract `_metagraph_hotkey_to_uid(metagraph)` as a module-level private
helper and call it from both functions so the map is always derived
from the same snapshot in one place.

## Scope

Pure refactor. No behavior change, no signature change.

Fixes #169

### DIFF
--- a/allways/commitments.py
+++ b/allways/commitments.py
@@ -119,6 +119,15 @@ def get_commitment(subtensor: bt.Subtensor, netuid: int, hotkey: str, block: Opt
     return decode_commitment_field(metadata)
 
 
+def _metagraph_hotkey_to_uid(metagraph: 'bt.Metagraph') -> dict:
+    """Build a {hotkey: uid} map from a metagraph snapshot.
+
+    Extracted so callers always derive the map from the same snapshot
+    and cannot accidentally use a stale copy.
+    """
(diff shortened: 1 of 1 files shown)


igeabdulrahmanikeoluwa-str:
@anderdc , on for review 🫡

anderdc:
I don't think this extraction actually helps much at all

PR #317, opened 2026-05-12
fix: force one scoring pass per process so mid-window restarts don't bleed emissions to RECYCLE

LandynDev:
## Problem → Solution

**Problem — restart mid-window leaves stale scores until next 1200-step boundary**
- `score_and_reward_miners` only fires on `self.step % SCORING_WINDOW_BLOCKS == 0` (`validator/forward.py:85`).
- `self.step` and `self.scores` persist across restarts via `state.npz` (`base/validator.py:289-298`).
- A restart landing at e.g. `step=791` skips scoring for ~409 more steps (~80–100 min) while `set_weights` keeps running every epoch on the stale/zero scores.
- Empty-norm fallback (`base/validator.py:174-183`) then routes 100% weight to `RECYCLE_UID` (53) on every emit during that gap.
- Observed after PR #310 (50/50 pool weights) deployed and validators restarted mid-window.
- → **Force one scoring pass per fresh process via a non-persisted boolean flag.** Steady-state 1200-step cadence preserved.

## Change

Two-file diff, ~5 lines:

- `neurons/validator.py` — `self.initial_scoring_done = False` next to existing per-run flags (not in `state.npz`).
- `validator/forward.py` — gate becomes `if not self.initial_scoring_done or self.step % SCORING_WINDOW_BLOCKS == 0:`, flips the flag after the call.

## Why the trigger lives in `forward()`, not after `load_state()`

`load_state` runs inside `BaseValidatorNeuron.__init__`, before `event_watcher.sync_to(self.block)` and `poll_commitments` — both feed `replay_crown_time_window`. Calling scoring straight after `load_state` would replay a stale window. Gating in `forward()` means the first scoring pass runs *after* event sync + commitment poll, so the window has fresh state. SQLite state.db carries rate_events + swap_outcomes across restart already.

## What this does NOT change

- `SCORING_WINDOW_BLOCKS` steady-state cadence (still 1200).
- `SCORING_EMA_ALPHA = 1.0`.
- `set_weights` cadence, epoch_length, or the empty-norm recycle fallback (kept as safety net for genuinely empty windows).
- PR #310 pool weights — already correct; this fix just ensures they get read on the next forward instead of in ~80 min.

## Test plan

- [x] `ruff format` / `ruff check allways/ neurons/` — clean
- [x] `pytest tests/` — 411/411 passing locally
- [ ] After merge: restart `aw-vali`, expect `forward: scoring done` on the **first** forward step (not 1200 later); next `set_weights` emits on the live miner UID, not UID 53; subsequent `forward: scoring done` only at `step % 1200 == 0`.

Plan: `plans/scoring-cadence-cold-start.md`.

### DIFF
--- a/allways/__init__.py
+++ b/allways/__init__.py
@@ -1,3 +1,3 @@
-__version__ = '1.0.1'
+__version__ = '1.0.2'
 version_split = __version__.split('.')
 __spec_version__ = (1000 * int(version_split[0])) + (10 * int(version_split[1])) + (1 * int(version_split[2]))
--- a/allways/constants.py
+++ b/allways/constants.py
@@ -49,7 +49,7 @@ BTC_FEE_RATE_SAFETY_MULTIPLIER = 1.25
 DEFAULT_MINER_TIMEOUT_CUSHION_BLOCKS = 5
 
 # ─── Scoring ─────────────────────────────────────────────
-SCORING_WINDOW_BLOCKS = 1200  # ~4 hours at 12s/block — also the scoring cadence
+SCORING_WINDOW_BLOCKS = 600  # ~2 hours at 12s/block — also the scoring cadence
 SCORING_EMA_ALPHA = 1.0  # Instantaneous — no smoothing across passes
 CREDIBILITY_WINDOW_BLOCKS = 216_000  # ~30 days
 DIRECTION_POOLS: dict[tuple[str, str], float] = {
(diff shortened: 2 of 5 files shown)


Helios531:
hi, @LandynDev, can you assign me on this issue, plz.

anderdc (approved):

PR #328, opened 2026-05-14
fix: refuse BTC source send when reservation TTL is too short to extend (#315)

### LINKED ISSUE #315: BTC source swaps can expire after funds are sent because 3-conf wait exceeds reservation TTL
## Summary

`alw swap now --from btc --to tao` can successfully reserve a miner and broadcast the BTC source transaction, then fail to initiate any on-chain swap because the reservation expires before the required BTC confirmations arrive.

In this case, BTC was sent to the selected miner address, the reservation later disappeared, `next_swap_id` did not advance, and no TAO was received.

This looks like a mismatch between:

- BTC source confirmation requirement: 3 confirmations, displayed by CLI as ~30 min
- reservation TTL: ~50 Bittensor blocks / ~10 min

The CLI warns not to send near the end of the reservation, but even sending immediately can miss the TTL when BTC requires 3 confirmations.

## Reproduction / observed flow

Quote:

```bash
alw swap quote --from btc --to tao --amount 0.00018
```

Selected route:

```text
UID 136
Rate: 1 BTC -> 255 TAO
Expected receive: 0.04544100 TAO after 1% fee
```

Swap command:

```bash
alw swap now \
  --from btc \
  --to tao \
  --amount 0.00018 \
  --receive-address 5E5PDWy3msrnA4uucy5bWRPVkFrzip4DTNN2qcEnPe5Tb24j \
  --from-address bc1qqnja4d8q0jxnv06v6f2sxj9570klg3hm46cnul \
  --auto \
  --btc-fee-rate 3
```

CLI output showed:

```text
Miner reserved! You have ~10 min to send your funds before the reservation expires.
...
BTC sent (tx: 51c83350cd3dc695f6a53bbfe0709aeb04cba5387fa47a44ffd790457e234c3a)
...
V1: queued Queued — 0/3 confirmations. Validator will auto-initiate when confirmed.
Waiting for 3 BTC confirmation(s) (~30 min).
```

## On-chain / transaction evidence

BTC tx:

```text
51c83350cd3dc695f6a53bbfe0709aeb04cba5387fa47a44ffd790457e234c3a
```

Mempool:

```text
https://mempool.space/tx/51c83350cd3dc695f6a53bbfe0709aeb04cba5387fa47a44ffd790457e234c3a
```

Confirmed output to selected miner BTC address:

```text
bc1qdpw3zrh50ygk8akrg69lnyt3q3qkkdwxw9pth2
18,000 sats / 0.00018 BTC
```

Change output back to source wallet:

```text
bc1qqnja4d8q0jxnv06v6f2sxj9570klg3hm46cnul
1,373 sats
```

Expected TAO receive address:

```text
5E5PDWy3msrnA4uucy5bWRPVkFrzip4DTNN2qcEnPe5Tb24j
```

Expected receive amount:

```text
0.04544100 TAO
```

## Current Allways state after BTC confirmation

```text
alw view reservation:
Reservation is no longer active on-chain.
Either the reservation expired before you sent funds, or your swap already initiated and has since completed.
Local reservation state cleared.
```

```text
alw view active-swaps:
No swaps found
```

```text
alw view swap 5:
Swap 5 does not exist. Next swap ID: 5.
```

Coldkey TAO balance did not increase.

## Expected behavior

The CLI/protocol should not allow a BTC source swap to enter a state where funds are sent but the reservation TTL is too short for the required confirmation policy.

Possible fixes:

- extend reservation TTL for BTC source swaps to cover the required confirmation window plus margin
- reduce BTC confirmation requirement or make it configurable by risk policy
- prevent auto-send when `required_confirmations * btc_block_time` exceeds remaining reservation TTL
- after BTC tx broadcast, extend/finalize reservation automatically until confirmation can initiate the swap
- surface a hard preflight error instead of a warning when the confirmation window cannot fit inside the reservation TTL

## Impact

This creates a user-funds-loss condition: BTC can be sent to the miner, but no swap is initiated and no TAO is delivered.

Khaostica:
## Summary

Add a CLI-side preflight that classifies remaining reservation TTL at send time and refuses to broadcast source funds into a reservation the validator auto-extension flow cannot rescue. Closes #315.

## Problem
`alw swap now --from btc --to tao` could successfully reserve a miner and broadcast the BTC source transaction, then fail to initiate any on-chain swap because the reservation expired before the required BTC confirmations arrived. Per #315, a user sent 0.00018 BTC against a 50-block (~10 min) reservation — BTC needs 2 confirmations (~20 min), so the reservation expired before the tx confirmed and no TAO was delivered.

#316 addressed two of the five proposed mitigations: it dropped BTC `min_confirmations` from 3 → 2 and added validator-side auto-extension that fires when the source tx becomes visible. But the auto-extension has a hard prerequisite — validators need at least `EXTEND_THRESHOLD_BLOCKS` (~18 blocks ≈ 3.6 min) of runway to land their propose tx and let its challenge window close (`allways/validator/optimistic_extensions.py:97`). Below that the propose is mechanically doomed and the reservation expires anyway. The CLI never checked this, so users could broadcast into a reservation no validator could rescue.

## What Changed
- Added `classify_send_runway()` in `allways/chains.py` — pure classifier returning `ok` / `extension_required` / `too_short` based on remaining TTL vs `EXTEND_THRESHOLD_BLOCKS` plus a 5-block (~1 min) propagation buffer.
- Added `preflight_send_runway()` in a new `allways/cli/preflight.py` module (placed outside `swap_commands/` so importing it doesn't trigger that package's eager command-registration `__init__.py` chain, which pulls in `bittensor` via `helpers`).
- Wired the gate into both `alw swap now` (after the "Send now?" confirmation) and `alw swap resume-reservation --send`, re-reading `current_block` at send time so an idle user at the summary panel doesn't bypass the check.
- Behavior by band:
  - `too_short`: hard refusal, suggests starting fresh — overrides `--yes` to avoid scripted callers silently losing funds.
  - `extension_required`: warns the validator auto-extension will need to kick in, requires explicit re-confirm (default no) in interactive mode, passes through in `--yes`.
  - `ok`: silent passthrough (no prompt, no print).

Implements solutions #3 ("prevent auto-send when required confirmations cannot fit inside reservation TTL") and #5 ("surface a hard preflight error instead of a warning") from #315.

## Tests
- `tests/test_chains.py::TestClassifySendRunway` — 7 cases covering the classifier's three bands, the BTC vs TAO confirmation-window difference, exact-floor boundary, zero remaining, and already-expired (negative remaining).
- `tests/test_preflight_send_runway.py` — 7 cases covering the wiring: subtensor re-read at send time, branch routing per band, and the `skip_confirm` contract (refuses on `too_short` even in `--yes`; passes through `extension_required` in `--yes`). The OK band asserts both `click.confirm` *and* `console.print` are untouched, so any future regression that adds chatter to the happy-path swap fails the test.

Verification:
```
pytest -q tests/test_chains.py tests/test_preflight_send_runway.py
```

### DIFF
--- a/allways/chains.py
+++ b/allways/chains.py
@@ -95,3 +95,49 @@ def compute_extension_target(
     # at MAX_EXTENSION_BLOCKS (lib.rs:670, :1090), so a deadline anchor blows
     # the cap whenever propose fires before the deadline.
     return current_subnet_block + blocks_needed
+
+
+# Status returned by ``classify_send_runway`` — see that function for semantics.
+RUNWAY_OK = 'ok'
+RUNWAY_EXTENSION_REQUIRED = 'extension_required'
+RUNWAY_TOO_SHORT = 'too_short'
+
+# Subtensor blocks set aside for the user's source tx to propagate to
+# validators' RPC view before the extension propose flow can pick it up.
+# ~1 minute covers typical Blockstream/Esplora indexing lag.
+SEND_PROPAGATION_BUFFER_BLOCKS = 5
+
+
+def classify_send_runway(
+    from_chain_id: str,
+    current_subnet_block: int,
+    reserved_until_block: int,
+    extend_threshold_blocks: int,
+) -> tuple[str, int]:
+    """Classify whether broadcasting the source tx now is safe.
+
+    The validator extension flow (optimistic_extensions.py) needs at least
+    ``extend_threshold_blocks`` of runway before the deadline to land a
+    propose tx and let its challenge window elapse — without that, the
+    propose is mechanically doomed and the reservation will expire. The
+    user's tx also needs a ~1 min propagation buffer so the validator can
+    actually see it before proposing.
+
+    Returns ``(status, remaining_blocks)`` where status is one of:
+      * ``RUNWAY_OK`` — full confirmation window fits in remaining TTL.
+        No validator extension needed; broadcast freely.
+      * ``RUNWAY_EXTENSION_REQUIRED`` — confirmations won't fit in TTL,
+        but there is enough runway for validators to auto-extend once
+        the tx is visible. Caller should warn but may proceed.
+      * ``RUNWAY_TOO_SHORT`` — TTL is below the extension floor; even
+        the auto-extension cannot rescue this broadcast. Caller should
+        refuse to send.
+    """
+    remaining = reserved_until_block - current_subnet_block
+    confirmation_subnet_blocks = confirmations_to_subtensor_blocks(from_chain_id)
+
+    if remaining < extend_threshold_blocks + SEND_PROPAGATION_BUFFER_BLOCKS:
+        return (RUNWAY_TOO_SHORT, remaining)
+    if remaining < confirmation_subnet_blocks + SEND_PROPAGATION_BUFFER_BLOCKS:
+        return (RUNWAY_EXTENSION_REQUIRED, remaining)
+    return (RUNWAY_OK, remaining)
--- a/allways/cli/preflight.py
+++ b/allways/cli/preflight.py
@@ -1,0 +1,79 @@
+"""Pre-broadcast safety gate for source funds.
+
+Sits outside ``swap_commands/`` so importing it does not trigger that
+package's eager command-registration ``__init__.py`` (which pulls in
+bittensor via helpers). That keeps the wiring unit-testable without the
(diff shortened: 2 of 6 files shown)


anderdc:
Closing in favor of #356, which takes a different organizational approach (no separate preflight module, cushion constants unified with the miner side, env override removed). The reservation-TTL safety for #315 will be a follow-up on top of #356.

PR #341, opened 2026-05-19
fix(miner): send BTC fulfillments from swap-pinned address

JSONbored:
## Summary
- send non-TAO destination funds from the `miner_to_address` pinned on the swap
- fail closed before broadcasting when a non-TAO swap is missing its pinned miner destination address
- add focused and invariant fulfillment tests for BTC sender selection and TAO unchanged behavior

## What changed
- `SwapFulfiller.send_dest_funds` now uses `swap.miner_to_address` as the BTC `from_address` hint
- stale or empty `my_addresses` cache values no longer determine the sender for active swap fulfillment
- non-TAO sends without a pinned sender return `None` before calling the provider
- table-driven invariant coverage now checks pinned BTC sender, stale cached sender, missing pinned sender, and unchanged TAO behavior together

## Why
Validators verify destination-chain fulfillment against the swap-pinned `miner_to_address`. If the miner falls back to wallet-wide BTC sender selection, it can broadcast funds from a different address and then fail validator verification despite sending the user funds.

## Validation
- `uv run pytest tests/test_fulfillment.py tests/test_bitcoin_signing.py -q`
- `uv run ruff check allways/miner/fulfillment.py tests/test_fulfillment.py`
- `uv run pytest -q`
- CodeRabbit review: skipped after retry because the CLI returned `rate_limit`

## Notes
- Checked for active upstream PR overlap before opening; no matching PR was found for this root cause.
- This is separate from the active double-broadcast PR: that PR protects crash windows after broadcast, while this one fixes which source address is used before broadcast.

### DIFF
--- a/allways/miner/fulfillment.py
+++ b/allways/miner/fulfillment.py
@@ -211,13 +211,16 @@ class SwapFulfiller:
             bt.logging.error(f'Swap {swap.id}: no provider for dest chain: {swap.to_chain}')
             return None
 
-        # Miner's own dest-chain sending address — cached from this miner's
-        # committed pair at startup (refreshed when the CLI signals a new
-        # post). For TAO we don't need a from_address hint because the wallet
-        # keypair fully identifies the sender. For non-TAO chains we pass the
-        # cached address so the provider can skip UTXO probing. Signing
-        # credentials live on the provider itself.
-        from_address = None if swap.to_chain == 'tao' else self.my_addresses.get(swap.to_chain)
+        # The validator verifies destination tx sender against the address
+        # snapshotted onto the swap, so non-TAO sends must use that same address.
+        from_address = None
+        if swap.to_chain != 'tao':
+            if not swap.miner_to_address:
+                bt.logging.error(
+                    f'Swap {swap.id}: missing miner_to_address for {swap.to_chain} fulfillment; refusing to send'
+                )
+                return None
+            from_address = swap.miner_to_address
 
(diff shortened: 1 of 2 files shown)


allways/miner/fulfillment.py
LandynDev, line 229: I understand the point of this PR, for the rare case where miner is flipping wallets, but still has an active swap to fulfill which uses their 'old' wallet.

Your updated `from_address` var gets passed here to `provider.send_amount()`. 
- For subtensor, this `from_address` is unused.
- For BTC, this `from_address` has its signing credentials fetched from `BTC_PRIVATE_KEY`. How would you properly distinguish at this lower level which signing credential to use? With these current code changes, it would pass `from_address` using the old address that got locked into the swap ✔️ but then it would sign with the newer wallet credentials since it has since been updated ❌ 

Overall, I'm not sure how much we should engineer this logic into the base miner because it unfolds to be more and more.
1) Miners should probably implement such custom logic in their own mining processes if they decide to take some type of heavy wallet rotating approach.
2) I kind of believe it's miners job to handle wallet rotating gracefully if that's what they decide to do and don't want to support it in the base miner provided. Yes, it's 'buggy', but it's not if miner just handles wallet rotations properly: `Deactivate miner -> Rotate wallets -> Reactivate miner`

LandynDev:
Check comment: https://github.com/entrius/allways/pull/341#discussion_r3277016959

PR #342, opened 2026-05-19
fix(validator): bind swapconfirm to reservation owner

### LINKED ISSUE #107: SwapConfirm allows reservation hijacking by a different user
### Problem
handle_swap_confirm() does not verify that the confirming from_address matches the from_addr stored in the active reservation.

The validator reads get_reservation_data(miner), but does not enforce that the reservation owner and the confirmer are the same address. Instead, it trusts the from_address supplied in the incoming SwapConfirm request and uses it for transaction verification and vote_initiate().

### Root Cause
The reservation already stores the user source address as from_addr, but handle_swap_confirm() never compares it against synapse.from_address.

Because that check is missing, confirmation is authorized by:
- active reservation exists for the miner
- submitted tx matches expected deposit address and amount
- submitted tx sender matches the caller-supplied from_address

It is not authorized by the address that actually created the reservation.

### Impact
A malicious user can take over another user’s active reservation.
A concrete attack flow is:
- User A reserves a miner.
- The reservation becomes active.
- User B sends the required source amount from User B’s own address.
- User B submits SwapConfirm with User B’s from_address, proof, tx hash, and payout address.
- Validators accept the confirm because the code never checks that the confirmer matches the reserved from_addr.
- Validators call vote_initiate() using User B’s addresses, even though User A created the reservation.

This breaks reservation ownership. An active reservation can be consumed by a different user than the one who reserved it, which is an authorization flaw and should be fixed urgently.

JSONbored:
## Summary
- Closes #107
- require `SwapConfirm` source addresses to match the active reservation owner before continuing
- preserve the upstream reservation-pin behavior while resolving the current `test` branch conflicts
- add regression coverage for owner mismatch, queue safety, and the existing queued-confirm path

## What changed
- `handle_swap_confirm` now reads the full reservation record and compares `reservation.from_addr` to `synapse.from_address`
- mismatched confirms return before reservation-pin resolution, proof verification, source transaction lookup, pending queue insertion, or contract voting
- tests cover missing full reservation data, mismatched source owners, no side effects on mismatch, and preservation of an existing SQLite pending row

## Why
A reservation is owned by the source address recorded on-chain. Confirming with a different source address should not be allowed to consume or modify that reservation, even if the caller controls the alternate address.

## Validation
- `uv run pre-commit run --all-files`
- `uv run pytest tests/test_axon_handlers.py -q`
- `uv run pytest tests/test_axon_handlers.py tests/test_swap_poller.py tests/test_swap_tracker.py -q`
- `uv run ruff check allways/validator/axon_handlers.py tests/test_axon_handlers.py`
- `uv run pytest -q`
- GitHub Actions: `pre-commit-fork` passed
- GitHub Actions: `pytest` passed
- CodeRabbit review against the uncommitted formatting fix: 0 issues
- CodeRabbit review against `upstream/test`: 0 issues

## Notes
- Rebased onto current upstream `test` to resolve conflicts from the merged reservation-pin and miner poller changes.
- Checked active upstream PRs before opening. Nearby validator PRs touch this area, but none appear to close #107 or add this owner-binding check.

### DIFF
--- a/allways/validator/axon_handlers.py
+++ b/allways/validator/axon_handlers.py
@@ -527,6 +527,14 @@ async def handle_swap_confirm(
 
             res_tao_amount, res_source_amount, res_dest_amount = res_data
 
+            reservation = contract.get_reservation(miner)
+            if reservation is None:
+                reject_synapse(synapse, 'Reservation record not found', ctx)
+                return synapse
+            if reservation.from_addr != synapse.from_address:
+                reject_synapse(synapse, 'Source address does not match active reservation', ctx)
+                return synapse
+
             # Prefer the commitment snapshot pinned when the user reserved.
             # A miner moving its rate or deposit address after the reservation
             # cannot then shortchange or rob the user — the swap settles
--- a/tests/test_axon_handlers.py
+++ b/tests/test_axon_handlers.py
@@ -12,11 +12,11 @@ import threading
 from unittest.mock import MagicMock, patch
 
 from allways.chain_providers.base import TransactionInfo
-from allways.classes import MinerPair
+from allways.classes import MinerPair, Reservation
 from allways.contract_client import ContractError
 from allways.synapses import SwapConfirmSynapse, SwapReserveSynapse
 from allways.validator.axon_handlers import handle_swap_confirm, handle_swap_reserve
-from allways.validator.state_store import ReservationPin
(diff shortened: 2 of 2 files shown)


anderdc (approved):

anderdc (changes requested):
FIx conflicts

JSONbored:
> FIx conflicts

Fixed.

anderdc (approved):

PR #344, opened 2026-05-19
fix: replace speculative resolve() with mark_voted() to prevent dupli…

### LINKED ISSUE #343: Bug: Duplicate swap votes on every confirmed/timed-out swap
### Summary

After a validator casts a successful confirm or timeout vote, the forward loop calls
`tracker.resolve()` immediately. `resolve()` discards the swap from `voted_ids` (line 86 
`swap_tracker.py`), destroying the duplicate-vote guard. On every subsequent forward step until
the `event_watcher` observes the on-chain `SwapCompleted`/`SwapTimedOut` event and calls
`resolve()` authoritatively, `is_voted()` returns `False` and a redundant extrinsic is
submitted for the same swap.

The correct alternative — `mark_voted()` — already exists, is tested, and was purpose-built for
this exact scenario: it adds to `voted_ids` without removing the swap from `active`, so
`is_voted()` stays `True` across steps until event-driven resolution clears it.

---

### Root cause

Two call sites substitute `tracker.resolve()` where `tracker.mark_voted()` is required:

| File | Line | Current (wrong) | Correct |
|---|---|---|---|
| `allways/validator/forward.py` | 408 | `tracker.resolve(swap.id, SwapStatus.COMPLETED, current_block)` | `tracker.mark_voted(swap.id)` |
| `allways/validator/forward.py` | 520 | `tracker.resolve(swap.id, SwapStatus.TIMED_OUT, self.block)` | `tracker.mark_voted(swap.id)` |

`resolve()` is intended to be called by `ContractEventWatcher` once the chain event is observed
— it is explicitly documented as the authoritative resolution path. Calling it speculatively from
the forward loop violates that ownership boundary and clears the guard prematurely.

---

### Reproduction

Every forward step between "our vote lands" and "event_watcher observes the finality event"
triggers a redundant extrinsic. On Bittensor, a forward step is ~12 s; the typical event lag is
several steps. The race fires on **every** successful swap confirm or timeout vote — not an edge
case.

---

### Impact

- **TAO fees:** Each redundant extrinsic burns fees. With multiple validators and a multi-step
  lag, the cost multiplies across the network.
- **Log noise:** "already voted" rejections from the contract flood the error log, masking
  legitimate failures.
- **Tally risk:** The contract likely deduplicates per voter, but a second vote arriving in a
  split-block window is an untested assumption. Any tally anomaly in that window would be
  difficult to diagnose.

---

### Fix

Replace the two speculative `resolve()` calls in `forward.py` with `mark_voted()`:

```python
# forward.py:408  (confirm_miner_fulfillments)
if voting.confirm_swap(self.contract_client, self.wallet, swap.id):
-   tracker.resolve(swap.id, SwapStatus.COMPLETED, current_block)
+   tracker.mark_voted(swap.id)
    bt.logging.success(f'Swap {swap.id}: verified complete, confirmed')

# forward.py:520  (enforce_swap_timeouts)
if voting.timeout_swap(self.contract_client, self.wallet, swap.id):
-   tracker.resolve(swap.id, SwapStatus.TIMED_OUT, self.block)
+   tracker.mark_voted(swap.id)
    bt.logging.warning(f'Swap {swap.id}: timed out')
```

`event_watcher` continues to call `tracker.resolve()` on `SwapCompleted`/`SwapTimedOut` events,
which is the correct authoritative path and will both clear `voted_ids` and drop the swap from
`active`.

---

### Testing

- `tests/test_swap_tracker.py` already covers `mark_voted` / `is_voted` / `prune_stale_voted_ids`
  round-trip behaviour.
- A forward-loop integration test should assert that after a successful `confirm_swap` or
  `timeout_swap` the swap is **not** re-submitted on the next step before `event_watcher` fires.

enjoyandlove:
### Summary

Closes   #343

- `tracker.resolve()` was being called speculatively in the forward loop after a confirm or timeout vote succeeded. `resolve()` removes the swap from `voted_ids`, destroying the duplicate-vote guard until `event_watcher` fires.
- Both call sites (line 408 `confirm_miner_fulfillments`, line 520 `enforce_swap_timeouts`) are replaced with `tracker.mark_voted()`, which records the vote without dropping the guard.
- `event_watcher` continues to call `tracker.resolve()` authoritatively on `SwapCompleted`/`SwapTimedOut` chain events — that ownership boundary is now respected.

### Why

Between "our vote lands" and "event_watcher observes the finality event" (~12 s per forward step, typically several steps of lag), `is_voted()` was returning `False` and a redundant extrinsic was submitted on every step. This burned TAO fees on every affected swap and flooded logs with "already voted" rejections, masking legitimate failures.

### Test plan

- [ ] Existing `tests/test_swap_tracker.py` `mark_voted` / `is_voted` / `prune_stale_voted_ids` round-trip tests pass
- [ ] `uv run pytest tests/`
- [ ] Verify that after a successful `confirm_swap` or `timeout_swap` the swap is not re-submitted on the next forward step before `event_watcher` fires

### DIFF
--- a/allways/miner/fulfillment.py
+++ b/allways/miner/fulfillment.py
@@ -12,6 +12,7 @@ from allways.chain_providers.base import ChainProvider, ProviderUnreachableError
 from allways.classes import Swap
 from allways.constants import DEFAULT_MINER_TIMEOUT_CUSHION_BLOCKS
 from allways.contract_client import AllwaysContractClient, ContractError, is_contract_rejection
+from allways.utils.logging import log_on_change
 from allways.utils.rate import expected_swap_amounts
 
 
@@ -191,7 +192,11 @@ class SwapFulfiller:
                 require_confirmed=True,
             )
             if tx_info is None:
-                bt.logging.debug(f'Swap {swap.id}: source tx not ready (not found, unconfirmed, or sender mismatch)')
+                log_on_change(
+                    f'src_waiting:{swap.id}',
+                    True,
+                    f'Swap {swap.id}: source tx not yet ready, will retry',
+                )
                 return False
 
             bt.logging.info(f'Swap {swap.id}: source funds verified ({tx_info.amount} from {tx_info.sender})')
@@ -271,7 +276,6 @@ class SwapFulfiller:
 
         # Step 2: Verify user sent source funds
         if not self.verify_user_sent_funds(swap, my_source_address):
-            bt.logging.debug(f'Swap {swap.id}: waiting for source funds confirmation')
             return False
 
         # Step 3: Send destination funds — unless we already did on a previous
--- a/allways/utils/logging.py
+++ b/allways/utils/logging.py
@@ -1,7 +1,7 @@
 import logging
 import os
 from logging.handlers import RotatingFileHandler
-from typing import Any
+from typing import Any, Optional
 
 import bittensor as bt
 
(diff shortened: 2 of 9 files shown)


enjoyandlove:
Hi, @anderdc 
I have just submitted this PR.When you have a moment, could you please review it?
Thank you for your time.

PR #367, opened 2026-05-22
feat: per-miner success rate in view rates and swap now

anderdc:
## What

Adds a `Reliability` column to `alw view rates` (the user-shopping table) and the `alw swap now` miner picker — per-miner, per-direction swap success rate, shown as `completed/resolved` and color-coded (green ≥90%, yellow ≥50%, red below).

## Why

Today a user picks a miner by posted rate alone. A miner can quote the best rate and still fail most of its swaps, with no signal in the CLI to catch it before reserving. The column surfaces it — e.g. a miner showing `T→B 1/17` has timed out 16 of its last 17 tao→btc swaps.

## How

- `fetch_miner_reliability()` (helpers.py) aggregates resolved swaps (`COMPLETED` + `TIMED_OUT` — the same basis as the subnet's `success_rate`) from the public swap-history API.
- 30-day window, matching the subnet's credibility window.
- Cached 10 min, keyed by API host + window; paging stops once swaps predate the window.
- Degrades gracefully — returns `None` / renders `—` if the API is unreachable, so `view rates` and `swap now` still work offline of the indexer.
- `ALLWAYS_API_URL` env override.

## Tests

`tests/test_miner_reliability.py` — aggregation, 30-day windowing, unresolved-status exclusion, API-unreachable, and a JSON-error-body guard.

## Docs

Companion docs PR in `allways-docs-ui` updates `cli.md`.

## Deferred (follow-ups)

- `--sort reliability` on `view rates`
- the same column on `alw view miners`
- making `--auto` reliability-aware

### DIFF
--- a/allways/cli/swap_commands/helpers.py
+++ b/allways/cli/swap_commands/helpers.py
@@ -117,7 +117,7 @@ def fetch_miner_reliability(window_days: int = RELIABILITY_WINDOW_DAYS, use_cach
         resolved = _parse_iso(s.get('resolvedAt')) or _parse_iso(s.get('updatedAt'))
         if resolved is None or resolved < cutoff:
             continue
-        direction = f"{s.get('sourceChain')}->{s.get('destChain')}"
+        direction = f'{s.get("sourceChain")}->{s.get("destChain")}'
         comp, tot = stats.setdefault(hk, {}).get(direction, (0, 0))
         stats[hk][direction] = (comp + (status == 'COMPLETED'), tot + 1)
 
--- a/allways/cli/swap_commands/helpers.py
+++ b/allways/cli/swap_commands/helpers.py
@@ -4,7 +4,6 @@ import sys
 import tempfile
 import time
 from dataclasses import asdict, dataclass
-from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Optional, Tuple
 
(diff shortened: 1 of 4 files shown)


LandynDev:
Pushed two commits to this branch to get it ready to land:

1. **Rebased onto `test`** — resolved the import-ordering conflict with #369 (`dashboard_url` vs `fetch_miner_reliability`).
2. **Moved the aggregation server-side.** Instead of paging the entire `/swaps` feed and re-deriving completed/total in the CLI, `fetch_miner_reliability()` now makes a single call to a new `GET /miners/reliability` endpoint that returns the per-direction counts pre-aggregated. That keeps the aggregation in the layer that owns it, drops the client-side windowing/status-filtering (now server-side), and shrinks the new API surface to one fast request — which matters since this is the CLI's first API dependency. The `None`-on-failure graceful-degradation contract is untouched: api down → dashes + 'unavailable' notice, CLI still fully functional.

Net effect on the Python side: −66 lines, same behavior.

**Coordinating PR:** entrius/das-allways#24 adds the endpoint. Merge + deploy that first; until it's live the column just renders `—` (graceful), so no hard breakage, but the data won't show until the API ships it.

Tests updated to the single-endpoint shape (5 pass); ruff clean.

entrius (approved):

PR #419, opened 2026-05-28
Vote deactivate active miners below min_collateral

### LINKED ISSUE #387: [Bug] Validators never call vote_deactivate after min_collateral raise — under-min miners stay active
## Duplicate check

| Item | Relation |
|------|----------|
| **#186** (closed) | `vote_deactivate` **quorum doesn't cancel in-flight rounds** — contract-side, not missing validator caller |
| **#177** (closed) | Auto-deactivation **partial state** on penalty path — different trigger |
| Scoring comment (scoring.py ~483–486) | Documents intent to use `vote_deactivate` for min-raise — **no implementation** |
| No open/closed issue | Validator never invokes `vote_deactivate` anywhere |

## Description

The contract exposes `vote_deactivate` for remediation when admin raises `min_collateral`. Existing active miners with collateral between the old and new minimum remain `active=true` — they cannot take new reservations (`vote_reserve` checks live min) but still qualify for crown replay while active. Validator code has a `contract_client.vote_deactivate` wrapper but **no caller** in the repo; scoring comments assume validators will use it for the min-raise edge case.

## Steps to Reproduce

1. Miner activates with 5 TAO collateral; `min_collateral = 1 TAO`.
2. Owner calls `set_min_collateral(10 TAO)`.
3. Miner stays `active=true` on-chain.
4. Grep codebase: `vote_deactivate(` only hits contract metadata and `contract_client.py` — no validator forward/scoring path calls it.

## Expected Behavior

Validators detect active miners below current `min_collateral` and submit `vote_deactivate` (quorum-gated), as the contract docs describe.

## Actual Behavior

Under-min miners remain active indefinitely until slashed or manually deactivated.

## Environment

- Branch: `test`
- Component: validator + contract integration

## Additional Context

**PR scope:** Add a forward or scoring pass: for each active miner, if `collateral < get_min_collateral()`, submit `vote_deactivate`.

kiannidev:
## Summary

Fixes [#387](https://github.com/entrius/allways/issues/387) — when admin raises `min_collateral`, existing active miners between the old and new floor stay `active=true` on-chain. Validators had `vote_deactivate` in `contract_client` but never called it.

## Related Issue

Fixes #387

## Change Type

- [x] Bug fix
- [x] Regression tests
- [ ] New feature
- [ ] Refactor

## What Changed

- Added `enforce_under_min_collateral_deactivation` forward pass after `poll_commitments`
- For each locally-tracked active miner still active on-chain: if `collateral < min_collateral`, submit `vote_deactivate`
- Added `voting.vote_deactivate` wrapper (mirrors `confirm_swap` / `timeout_swap`)

## Files Changed

| File | Change |
|------|--------|
| `allways/validator/forward.py` | New deactivation pass |
| `allways/validator/voting.py` | `vote_deactivate` helper |
| `tests/test_under_min_collateral_deactivation.py` | Regression tests |

## Validation

```bash
cd /root/gittensor/allways
uv run pytest tests/test_under_min_collateral_deactivation.py -v
```

## Test Plan

- [x] Under-min active miner → `vote_deactivate` called
- [x] At/above floor → no vote
- [x] `min_collateral == 0` → skip
- [x] Contract-inactive miner skipped even if in local active set
- [ ] CI green on `test`

### DIFF
--- a/allways/validator/forward.py
+++ b/allways/validator/forward.py
@@ -82,6 +82,8 @@ async def forward(self: Validator) -> None:
 
     poll_commitments(self)
 
+    enforce_under_min_collateral_deactivation(self)
+
     # Pull newly-initiated and resolved swaps off the contract.
     await tracker.poll()
     bt.logging.info('forward: tracker polled')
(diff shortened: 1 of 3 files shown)


anderdc:
Please estimate the runtime overhead per forward block for adding this. Each forward step for the validator is extremely sensitive and if we do this every single block it will affect runtime

kiannidev:
> Please estimate the runtime overhead per forward block for adding this. Each forward step for the validator is extremely sensitive and if we do this every single block it will affect runtime

Here's an estimate of the current implementation:

**Per forward step (~every 12s block):**
- `bounds_cache.min_collateral()` — cached with a 300-block TTL, so amortized RPC cost is ~0
- Loop over `event_watcher.active_miners` (N = active miners, usually small):
  - `get_miner_active_flag(hotkey)` → 1 contract RPC
  - `get_miner_collateral(hotkey)` → 1 contract RPC

So steady-state cost is **~2N contract reads per block**, even when every miner is above the floor. That's heavier than this edge case needs — `min_collateral` raises are rare, and remediation doesn't need sub-block latency.

**Planned adjustment in this PR:**
1. Run the pass on a reduced cadence (e.g. when the `min_collateral` cache refreshes / ~every 300 blocks, not every block)
2. Drop the redundant `get_miner_active_flag` check — `event_watcher.active_miners` already reflects the active set
3. Use a single `get_miner_snapshot` per candidate instead of two separate reads

That should make normal operation negligible while still catching under-min miners within minutes of a floor raise. I can push that follow-up here if that cadence works for you.

PR #449, opened 2026-06-04
feat(scoring): crown-dominant quality-weighted-volume blend with a single credibility term

anderdc:
## What & why

Reshapes the reward from a crown-time base with bolted-on correction factors into a **crown-dominant blend of availability and realized service**, and collapses the two reliability terms into one. Builds on the original depth-quality work (the self-referential market reference, retained below).

```
reward = pool · credibility · (λ·qvol_share + (1−λ)·crown_share·cap·quality)
```

- **`qvol_share`** — the miner's share of the network's **quality-weighted swap volume** (per direction): each completed swap's TAO leg scaled by `quality_factor(clearing_rate)`. This is the *service* half — real TAO moved, at the rate it cleared. Volume now **earns** (a non-crown server collects the λ slice) instead of merely gating crown.
- **`crown_share·cap·quality`** — the *availability* half: best-rate hold time, scaled by capacity and the per-block depth quality (the original 449 weighting, retained).
- **`λ = QVOL_REWARD_WEIGHT = 0.3`** — **crown-DOMINANT on purpose.** Realized volume is sybil/wash-inflatable until swap counterparties are verifiable, so the volume share is kept small. **Raising λ toward volume is gated on wash filtering** — documented at the constant. Going volume-dominant before wash is filtered would entrench wash farms (winner-take-all crown protects honest rate competition better than volume-share does against fakeable volume).
- **Adaptive λ** — with no realized volume (bootstrap) crown takes the full pool; with volume but no crown holder, volume takes it. An empty component never silently recycles its share.

## The depth reference (from the original feature, unchanged)

The `quality_factor` is anchored to a **self-referential per-direction market reference**: a trimmed, volume- and recency-weighted (EMA half-life) average of the subnet's own *completed-swap* clearing rates. No oracle, no contract change. Recomputed each round from the stored `clearing_rate` column — no persisted EMA state, so every validator derives a byte-identical value (stable total-order sort + `math.fsum`, `now_block = window_end`). Trim + volume-weight + per-miner cap defend it against single-actor manipulation. Disabled below `QUALITY_N_MIN` (factor 1.0) so it's a guaranteed no-op until real swap history accrues.

`quality_factor ∈ [QUALITY_FLOOR, 1.0]`: floor at/below market, full at `QUALITY_ANCHOR` (5%) deeper, linear between; never below the floor.

## Reliability: one term, not two

Merges `success_rate³ × credibility_ramp` into a single **`credibility`** (the two double-counted timeouts). Drops `SUCCESS_EXPONENT` and the `success_rate` helper. The `>2`-timeout cliff is preserved.

⚠️ **Known follow-up:** the timeout cliff is an *absolute count*, so it's volume-unaware — a high-volume miner is killed by 3 timeouts (~0.3% fail) while a 3-swap/2-timeout miner is fully credible. Fine while crown dominates; should be **volume-scaled before λ grows**. Flagged in the `credibility` docstring.

## Changes

- **`scoring.py`** — `quality_weighted_volume` helper; per-direction qvol + adaptive-λ blend in `calculate_miner_rewards`; `credibility` replaces `success_rate`/`SUCCESS_EXPONENT`/`volume_factor`/`VOLUME_WEIGHT_ALPHA`.
- **`constants.py`** — `QVOL_REWARD_WEIGHT = 0.3` (with the wash-gating note); removed `SUCCESS_EXPONENT`, `VOLUME_WEIGHT_ALPHA`.
- **`scoring_trace.py`** — trace surfaces `crown_share`, `qvol_share`, `credibility`, `quality`; `success_rate`→`credibility` plumbing.
- **`state_store.py` / `event_watcher.py`** — the depth feature's `clearing_rate` column (+ idempotent migration), per-direction query, and `SwapCompleted` population (unchanged from the original feature).

## Scope / not in this PR

**No wash filtering.** `swap_outcomes` records no counterparty, so self-dealing volume is unfilterable from the event stream — that needs a contract change to emit the depositor. The crown-dominant λ is the interim guard. Wash-resistance + a volume-dominant λ are the v2 direction.

## ⚠️ Calibration

`QUALITY_ANCHOR` (0.05) is a placeholder — **calibrate against live rate dispersion before trusting it**. `QVOL_REWARD_WEIGHT` (0.3) is a deliberately conservative starting point.

## Tests

Reference math (incl. shuffle-invariance), `quality_factor`/`quality_weighted_volume`/`credibility` unit tests, and e2e blend/credibility/capacity/quality weighting via `calculate_miner_rewards`. Full suite green; ruff clean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,8 @@ tmp/
 data/
 
 CLAUDE.md
+
+# Local tooling / scratch
+.mcp.json
+*.md
+.vouch/
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ __pycache__
 CLAUDE.md
 .claude/
 .vscode/
+.mcp.json
+.vouch
+*.md
 
 # Rust/Cargo build artifacts
 target/
--- a/allways/__init__.py
+++ b/allways/__init__.py
@@ -1,3 +1,3 @@
-__version__ = '1.0.10'
+__version__ = '1.0.11'
 version_split = __version__.split('.')
 __spec_version__ = (1000 * int(version_split[0])) + (10 * int(version_split[1])) + (1 * int(version_split[2]))
--- a/allways/chain_providers/base.py
+++ b/allways/chain_providers/base.py
@@ -31,6 +31,11 @@ class ChainProvider(ABC):
     3. Add {ENV_PREFIX}_* vars to .env
     """
 
+    # True if this provider's RPCs hit the shared substrate websocket, so
+    # callers must serialise them under axon_lock. HTTP-backed providers leave
+    # this False and stay lock-free.
+    uses_substrate: bool = False
+
     @abstractmethod
     def get_chain(self) -> ChainDefinition: ...
 
--- a/allways/chain_providers/bitcoin.py
+++ b/allways/chain_providers/bitcoin.py
@@ -4,10 +4,11 @@ from typing import Any, Optional, Tuple
 from urllib.parse import urlparse
 
 import base58
-import bech32
 import bittensor as bt
 import requests
 from bitcoin_message_tool.bmt import sign_message, verify_message
+from embit.networks import NETWORKS
+from embit.script import address_to_scriptpubkey
 
 from allways.chain_providers.base import ChainProvider, ProviderUnreachableError, TransactionInfo
 from allways.chains import CHAIN_BTC, ChainDefinition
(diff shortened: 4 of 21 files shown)


JSONbored:
Reading the "no wash filtering — needs a contract change to emit the depositor" note — totally fair for self-dealt *volume*. But I had a thought on one slice that might not need the counterparty at all, since the settlement addresses are already in every commitment.

If I'm reading the reward loop right, `crown_share` and `qvol_share` accrue per hotkey with no grouping by settlement identity, and the per-miner cap only bounds weight in the quality *reference*, not the reward — so hotkeys sharing one BTC+TAO settlement address would still split the pool as N independent miners off one pool? `λ=0.3` damps the volume half; I just wasn't sure anything caps the crown half (possible I'm missing it).

Why it might matter in practice: on SN7 right now, reading the Commitments pallet, one operator is running ~6 coldkeys that all post to just 3 BTC settlement addresses (one address backs the original coldkey plus a newer one; another backs three), holding the BTC→TAO crown above fair via self-reservation. So coldkey-level identity is already being churned — the settlement tuple is the only stable handle. (One of those addresses is also the one from #444's self-send write-up — A→A that's grown into A-across-N-coldkeys.)

Idea, fwiw: collapse active hotkeys sharing a canonical settlement tuple into one identity before the crown/qvol split — deterministic, no counterparty or contract change, in the spirit of #444 and robust to the coldkey-churn above. Could be a cheap interim complement to `λ`, or maybe that's already where v2 is headed. Happy to share the coldkeys/addresses + a failing test if useful, or I'm also happy to draft up an issue for this.

Issue #448, opened 2026-06-03
[Bug] is_valid_address rejects all Taproot (P2TR / bech32m) addresses — TAO→BTC swaps to a bc1p… payout address always fail


cleanjunc:
## Summary

`BitcoinProvider.is_valid_address` rejects **every** valid Taproot (P2TR) Bitcoin address. As a result, any **TAO→BTC** swap whose BTC receiving address is a Taproot address (`bc1p…` on mainnet, `tb1p…`/`bcrt1p…` on test networks) is rejected 100% of the time — both in the CLI before sending and in the validator at confirm time — with a misleading "Invalid destination address format" error.

Taproot is the default receive-address type in many modern wallets and exchanges, so this blocks a large and growing class of legitimate swaps.

## Root cause

`allways/chain_providers/bitcoin.py` (`is_valid_address`):

```python
def is_valid_address(self, address: str) -> bool:
    """Validate BTC address format without RPC (bech32/base58 decode)."""
    if not address or not isinstance(address, str):
        return False
    try:
        if address.lower().startswith(('bc1', 'tb1', 'bcrt1')):
            hrp, data = bech32.bech32_decode(address)
            return data is not None
        decoded = base58.b58decode_check(address)
        return len(decoded) == 21 and decoded[0] in (0x00, 0x05, 0x6F, 0xC4)
    except Exception:
        return False
```

`bech32.bech32_decode` validates **only** the BIP-173 *bech32* checksum (polymod constant `1`). Taproot addresses (witness v1) are encoded with *bech32m* (BIP-350, polymod constant `0x2bc830a3`), so `bech32_decode` returns `(None, None)` for them and the function returns `False`.

## Reproduction

Run against the pinned `bech32` dependency:

```python
import bech32, base58

def is_valid_address(address):
    if not address or not isinstance(address, str):
        return False
    try:
        if address.lower().startswith(('bc1', 'tb1', 'bcrt1')):
            hrp, data = bech32.bech32_decode(address)
            return data is not None
        decoded = base58.b58decode_check(address)
        return len(decoded) == 21 and decoded[0] in (0x00, 0x05, 0x6F, 0xC4)
    except Exception:
        return False

cases = {
    'P2WPKH v0 (bc1q)': 'bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4',
[...]

PR #463, opened 2026-06-09
fix(bitcoin): enforce network in is_valid_address to reject wrong-network addresses

### LINKED ISSUE #460: enhancement: `is_valid_address` should reject BTC addresses from a different network than `BTC_NETWORK`
## Summary

`BitcoinProvider.is_valid_address` validates an address's *format* but not its *network*. A provider configured for mainnet (`BTC_NETWORK=mainnet`, or the lightweight default) accepts testnet (`tb1q…`, `m…`, `n…`), testnet4, and regtest (`bcrt1q…`) addresses as "valid." This check at validator confirm and at the CLI pre-check is the only gate that runs **before** the user's source funds are locked and **before** a miner moves real funds. Because it passes a wrong-network destination, the swap is initiated, the miner sends real BTC to the scriptPubKey decoded from that address (which re-encodes to a *different* address on the live network), and then the validator's destination verification can never match it — stranding the miner's funds.

## Current behavior

`allways/chain_providers/bitcoin.py`:

```python
def is_valid_address(self, address: str) -> bool:
    """Validate BTC address format without RPC (embit decode; all types incl. Taproot)."""
    if not address or not isinstance(address, str):
        return False
    try:
        return address_to_scriptpubkey(address) is not None
    except Exception:
        return False
```

`address_to_scriptpubkey` decodes addresses of any network, so this returns `True` regardless of whether the address matches `self.network`. `is_valid_address` never references `self.network`.

The miner's fulfillment path has **no network guard**. `send_dest_funds` calls `provider.send_amount(swap.user_to_address, …)` directly (`allways/miner/fulfillment.py`), and `send_amount_lightweight` builds the output straight from the user string:

```python
to_script = address_to_scriptpubkey(to_address)   # send_amount_lightweight
...
tx.vout.append(TransactionOutput(amount, to_script))
```

For a testnet/regtest `to_address` this produces the scriptPubKey whose witness program / hash160 is shared with the mainnet equivalent, and the tx is broadcast on the live (mainnet) network. The funds land at the mainnet re-encoding the user never saw.

(Pre-existing, not introduced by #453 — the prior hand-rolled validator also accepted testnet base58 version bytes and `tb1`/`bcrt1` bech32 prefixes. #453 only changed the decoder.)

## What actually happens downstream (corrected)

There **is** a downstream check, but it is an exact-string recipient match, not a network check, and it fires only **after** the miner has already sent real funds:

- The validator verifies the destination tx with `expected_recipient = swap.user_to_address` (`allways/validator/chain_verification.py`), i.e. the original `tb1q…` string the user entered.
- `api_verify_transaction` compares `vout.scriptpubkey_address == expected_recipient` with **no mainnet normalization** (`to_mainnet_address` is only used in the proof-signing path, not in tx verification).
- A mainnet Esplora returns the re-encoded `bc1q…` for that vout, so `bc1q… == tb1q…` is `False` → **destination verification fails.**

So this is not a silent, successfully-completed misdirected swap. The concrete failure mode is: the miner sends real BTC to the re-encoded address, the validator can never match the tx back to the swap, and the swap does not confirm. The miner is out real funds with no on-chain credit — the code already anticipates this state with a warning ("funds may have been sent without on-chain credit", `cleanup_stale_sends` in `fulfillment.py`).

## Why it matters

The destination address is user-entered for a TAO→BTC swap. A wrong-network paste (e.g. a testnet address from the user's wallet on a mainnet deployment) currently:

1. Passes the CLI pre-check (`swap.py`) and validator confirm (`handle_swap_confirm`'s `to_provider.is_valid_address(synapse.to_address)`) — format only.
2. Locks the user's source funds and reserves a miner.
3. Has the miner broadcast real BTC to `address_to_scriptpubkey(wrong_net_address)`, whose scriptPubKey re-encodes to a different address on the live network.
4. Fails the validator's exact-string destination match, so the swap never confirms.

The cheapest, correct place to stop this is the `is_valid_address` gate at step 1, before any funds move. The misdirected BTC may be recoverable when the address came from the user's own wallet (the re-encoded mainnet address shares the same key), but only if the miner notices, identifies the re-encoded address, and can spend it; when the wrong-network address belongs to a third party the user does not hold the key for, the funds are unrecoverable. Either way the right outcome is to reject the swap up front with a clear "that's a testnet address" message.

## Reproduction

**Unit level — directly reproducible, no node or funds:**

```python
from embit.script import address_to_scriptpubkey
from embit.networks import NETWORKS

# A mainnet-configured provider validating these destinations:
for a in [
    'tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx',   # testnet P2WPKH
    'mipcBbFg9gMiCh81Kj8tqqdgoZub1ZJRfn',           # testnet P2PKH
    'bcrt1qw508d6qejxtdg4y5r3zarvary0c5xw7kygt080', # regtest P2WPKH
]:
    spk = address_to_scriptpubkey(a)
    print(a, '-> valid:', spk is not None, '-> mainnet re-encode:', spk.address(NETWORKS['main']))
```

Output:

```
tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx -> valid: True -> mainnet re-encode: bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4
mipcBbFg9gMiCh81Kj8tqqdgoZub1ZJRfn         -> valid: True -> mainnet re-encode: 14JetYAhLevTRaePcAAX1vRMwaJt2QGRzp
bcrt1qw508d6qejxtdg4y5r3zarvary0c5xw7kygt080 -> valid: True -> mainnet re-encode: bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4
```

`is_valid_address` (which is just `address_to_scriptpubkey(addr) is not None`) returns `True` for all three on a mainnet provider.

**End-to-end (miner send → failed verification): not reproduced here.** That requires a live miner, validator, real BTC, and a live broadcast. The send-and-strand path above was confirmed by reading the code (`fulfillment.py` → `send_amount_lightweight` → `chain_verification.py`), not by running a swap. Treat it as a code-traced consequence, not a reproduced run.

## Suggested fix

Enforce the network in `is_valid_address`: detect the address's network from its prefix/version (the codebase already does prefix detection in `detect_address_type`) and compare to `self.network`, rejecting on mismatch. Format validation (including Taproot/bech32m via embit) stays as-is for the matching-network case.

Network → prefixes:

- mainnet: `bc1…`, `1…`, `3…`
- testnet / testnet4: `tb1…`, `m…`, `n…`, `2…`
- regtest: `bcrt1…`, plus the testnet base58 version bytes (`m…`, `n…`, `2…`)

Implementation notes:

- Treat `testnet` and `testnet4` as one network class for matching — they share address prefixes.
- Be deliberate about regtest base58: Bitcoin Core regtest reuses testnet's base58 version bytes, so a regtest provider must accept `m`/`n`/`2` as well as `bcrt1…`.

## Tests

Unit-testable as red/green, no judgment call:

- On a `BTC_NETWORK=mainnet` provider, assert `is_valid_address('tb1q…')`, `is_valid_address('mipc…')`, `is_valid_address('bcrt1q…')` return `False`, while a valid mainnet address returns `True`.
- Mirror for a testnet-configured provider.

**Note — this is not purely additive.** The existing `tests/test_bitcoin_signing.py::test_accepts_all_standard_types` builds a default (mainnet) lightweight provider and its `VALID` list includes `tb1p…`, `bcrt1q…`, and `bcrt1p…`, asserting they are accepted. That test encodes the current accept-everything behavior and must be updated to split addresses by network, not just extended.

## Scope / risk

- Small, self-contained change to one function, plus updating one existing test and adding new ones.
- Tightens an accept that should never have passed; the only addresses that flip accepted→rejected are wrong-network ones, which cannot be fulfilled correctly at the intended address anyway.

## Environment

- File: `allways/chain_providers/bitcoin.py` (`is_valid_address`).
- Affected flows: CLI swap pre-check (`allways/cli/swap_commands/swap.py`) and validator `handle_swap_confirm` destination validation (`allways/validator/axon_handlers.py`), TAO→BTC. Miner send path with no network guard: `allways/miner/fulfillment.py` → `send_amount_lightweight`. Downstream string-match verification: `allways/validator/chain_verification.py` → `api_verify_transaction`.

jaso0n0818:
## Problem

`BitcoinProvider.is_valid_address` validated address *format* (embit scriptpubkey decode) but never checked the address *network*. A mainnet-configured provider accepted testnet (`tb1q…`, `m…`, `n…`) and regtest (`bcrt1q…`) addresses as valid.

Failure mode (issue #460): a user pastes a wrong-network destination address → passes the CLI pre-check and validator confirm → miner broadcasts real BTC to the re-encoded scriptpubkey on the live network → validator's exact-string destination match fails → swap never confirms and miner loses funds with no on-chain credit.

## Fix

****

Add a module-level helper `_address_network(address)` that classifies an address into one of four disjoint network classes:

| Result | Covers |
|---|---|
| `mainnet` | `bc1q…`, `bc1p…`, `1…`, `3…` |
| `testnet-bech32` | `tb1q…`, `tb1p…` |
| `testnet-base58` | `m…`, `n…`, `2…` (shared version bytes, see below) |
| `regtest` | `bcrt1q…`, `bcrt1p…` |

Update `is_valid_address` to call it after the embit format check and accept only addresses whose class matches `self.network`:

- **mainnet** → `mainnet`
- **testnet / testnet4** → `testnet-bech32` or `testnet-base58`
- **regtest** → `regtest` or `testnet-base58`

The four-way split (rather than three-way) handles Bitcoin Core regtest correctly: regtest reuses testnet base58 version bytes (`m`/`n`/`2`) for legacy address types, so a regtest provider must accept those while still rejecting testnet-specific `tb1…` bech32 addresses.

## Tests

****

Replace the old `VALID` list (which mixed mainnet, testnet, and regtest addresses and tested all against a mainnet provider) with per-network address lists and network-specific tests:

- Mainnet provider: accepts mainnet, rejects testnet and regtest bech32
- Testnet provider: accepts tb1/m/n/2, rejects mainnet and regtest bech32
- Regtest provider: accepts bcrt1/m/n/2, rejects mainnet and testnet bech32

12 tests total, all pass. The #448 Taproot regression guard is preserved as `test_mainnet_accepts_taproot`.

Closes #460

### DIFF
--- a/allways/chain_providers/bitcoin.py
+++ b/allways/chain_providers/bitcoin.py
@@ -45,6 +45,29 @@ def detect_address_type(address: str) -> str:
     return 'unknown'
 
 
+def _address_network(address: str) -> str:
+    """Classify an address by its implied BTC network.
+
+    Returns one of:
+    - 'mainnet'         — bc1q…, bc1p…, 1…, 3…
+    - 'testnet-bech32'  — tb1q…, tb1p…  (testnet/testnet4 native-segwit only)
+    - 'testnet-base58'  — m…, n…, 2…  (testnet base58 version bytes,
+                          shared with regtest due to Bitcoin Core's design)
+    - 'regtest'         — bcrt1q…, bcrt1p…
+    - 'unknown'
+    """
+    if address.startswith(('bc1q', 'bc1p', '1', '3')):
+        return 'mainnet'
+    if address.startswith(('tb1q', 'tb1p')):
+        return 'testnet-bech32'
+    if address.startswith(('bcrt1q', 'bcrt1p')):
+        return 'regtest'
+    # base58 version bytes 0x6f (m/n) and 0xc4 (2) are shared by testnet and regtest
+    if address[:1] in ('m', 'n', '2'):
+        return 'testnet-base58'
+    return 'unknown'
+
+
 def to_mainnet_wif(wif: str) -> str:
     """Convert a testnet/regtest WIF (0xef) to mainnet (0x80) for signing libraries."""
     decoded = base58.b58decode_check(wif)
@@ -526,13 +549,31 @@ class BitcoinProvider(ChainProvider):
             return 0
 
     def is_valid_address(self, address: str) -> bool:
-        """Validate BTC address format without RPC (embit decode; all types incl. Taproot)."""
+        """Validate BTC address format and network against the configured BTC_NETWORK.
+
+        Rejects addresses that belong to a different network (e.g. testnet addresses
+        on a mainnet provider) to prevent misdirected on-chain sends.
+        """
         if not address or not isinstance(address, str):
             return False
         try:
-            return address_to_scriptpubkey(address) is not None
+            if address_to_scriptpubkey(address) is None:
+                return False
         except Exception:
             return False
+        addr_net = _address_network(address)
+        if addr_net == 'unknown':
+            return False
+        configured = self.network  # 'mainnet', 'testnet', 'testnet4', or 'regtest'
+        if configured in ('mainnet', ''):
+            return addr_net == 'mainnet'
+        if configured in ('testnet', 'testnet4'):
+            # tb1… (bech32) and m/n/2 (base58) are valid; bcrt1… is regtest-only
+            return addr_net in ('testnet-bech32', 'testnet-base58')
+        if configured == 'regtest':
+            # regtest uses bcrt1… bech32 and the testnet base58 version bytes (m/n/2)
+            return addr_net in ('regtest', 'testnet-base58')
+        return False
 
     def get_wif(self, address: str) -> Optional[str]:
         """Get WIF private key from env var or Bitcoin Core wallet."""
--- a/tests/test_bitcoin_signing.py
+++ b/tests/test_bitcoin_signing.py
@@ -274,30 +274,41 @@ class TestBroadcastedTxidsTracking:
         assert provider.broadcasted_txids == set()
 
 
-class TestIsValidAddress:
-    """BTC address format validation — must accept every standard type
-    (legacy, segwit v0, and Taproot/bech32m) and reject malformed input.
+def make_lightweight_provider_for_network(network: str) -> 'BitcoinProvider':
+    env = {'BTC_MODE': 'lightweight', 'BTC_PRIVATE_KEY': TEST_WIF, 'BTC_NETWORK': network}
+    with patch.dict(os.environ, env, clear=False):
+        return BitcoinProvider()
 
-    Taproot regression guard: the old bech32-only check rejected all `bc1p…`
-    addresses, blocking TAO->BTC payouts to Taproot wallets (issue #448).
+
+class TestIsValidAddress:
+    """BTC address format and network validation.
+
+    Each provider network must accept only its own addresses and reject
+    addresses from other networks (issue #460). Taproot regression guard:
+    the old bech32-only check rejected all bc1p… addresses (issue #448).
     """
 
-    # mainnet / testnet / regtest × P2PKH / P2SH / P2WPKH / P2WSH / P2TR
-    VALID = [
-        '1BvBMSEYstWetqTFn5Au4m4GFg7xJaNVN2',  # mainnet P2PKH
-        '3J98t1WpEZ73CNmQviecrnyiWrnqRhWNLy',  # mainnet P2SH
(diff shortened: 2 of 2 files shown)


jaso0n0818:
Rebased on latest `test`, local bitcoin signing tests pass (43/43). Small bugfix — ready for review.

PR #464, opened 2026-06-09
fix(bitcoin): size BTC fee-estimation inputs per address type (P2SH-P2WPKH 91 vB not 68 vB)

### LINKED ISSUE #459: fix(bitcoin): size BTC fee-estimation inputs per address type (P2SH-P2WPKH undersized by ~23 vB/input)
## Summary

`select_utxos` estimates the BTC network fee with a single segwit input size (68 vB) for *both* native and nested segwit. A P2SH-P2WPKH (nested segwit, `3…`) input is actually ~91 vB, so every nested-segwit input is undersized by ~23 vB. For a miner whose committed BTC address is P2SH-P2WPKH, multi-input sends can pay below the targeted fee rate — under fee pressure that means slower-than-targeted confirmation, risking a dest tx that misses the swap reservation window.

## Where

- `allways/chain_providers/bitcoin.py:791` — flat `input_vsize`:

```python
def select_utxos(self, utxos, amount: int, is_segwit: bool, fee_rate_override=None):
    fee_rate = self.estimate_fee_rate(override=fee_rate_override)
    input_vsize = 68 if is_segwit else 148
    ...
    est_vsize = 11 + len(selected) * input_vsize + 2 * 31
```

- `allways/chain_providers/bitcoin.py:663` — nested segwit collapsed into the 68 case:

```python
is_segwit = addr_type in ('p2wpkh', 'p2sh-p2wpkh')
```

The same function already attaches a `redeem_script` for `p2sh-p2wpkh` at `bitcoin.py:693-695` — i.e. the code knows about the 23-byte redeemScript push in the (×4-weighted) scriptSig that the fee estimate ignores.

- The bug is **duplicated** in `scripts/test_embit_send.py:70-71,80` (same `input_vsize = 68 if is_segwit else 148`). A fix should touch both sites.

## Why it's real (verifiable arithmetic, not speculation)

Input vsizes are fixed by the transaction format. Deriving from weight units:

| Input type   | non-witness   | witness | weight | vsize  |
|--------------|---------------|---------|--------|--------|
| P2WPKH       | 41 B ×4 = 164 | 108     | 272    | **68** |
| P2SH-P2WPKH  | 64 B ×4 = 256 | 108     | 364    | **91** |
| P2PKH        | —             | —       | —      | ~148   |

The extra ~23 vB on P2SH-P2WPKH is the 22-byte redeemScript (`OP_0 <20-byte hash>`) plus its 1-byte push, carried in the non-witness scriptSig.

### The 1.25× safety pad masks it for few inputs, then erodes

`est_vsize = 11 + n·input_vsize + 2·31` → `73 + 68n` for segwit. Holding outputs fixed to isolate the input error, the real tx is `73 + 91n`. At the true 1.25× multiplier (the high-fee / congested regime — exactly when this matters):

| Inputs (P2SH-P2WPKH) | est | real | est×1.25 ≥ real? |
|----------------------|-----|------|------------------|
| 1                    | 141 | 164  | 176 ✓            |
| 2                    | 209 | 255  | 261 ✓ (thin)     |
| 3                    | 277 | 346  | 346 (edge)       |
| 4+                   | 345 | 437  | 431 ✗            |

Solving `(73+91n) > 1.25·(73+68n)` gives the crossover at **n ≥ 4**. Multi-input sends happen exactly when no single UTXO covers the amount, so this path is reachable (e.g. a miner accumulating many small swap payments).

**Note on the multiplier:** `estimate_fee_rate` returns `int(round(base × 1.25))` — an integer sat/vB. The table above is the continuous-rate idealization; at high fee rates it's accurate (multiplier ≈ 1.25, threshold n≥4). At the fee floor (base≈5 → rate 6 → effective 1.2×) underpayment starts earlier (n≥2) but block space is plentiful there, so it's low-stakes.

### The override path has *no* pad at all

`estimate_fee_rate(override=…)` returns `max(1, override)` with **no 1.25 multiplier** (`bitcoin.py:854-855`), and `send_amount` threads `fee_rate_override` into `select_utxos` (`bitcoin.py:666`). So a manual/CPFP bump uses 68-vB sizing with **zero** safety margin — every nested-segwit input is undersized by the full 23 vB regardless of input count. This is the cleanest demonstration of the defect.

## Secondary: output sizing

The `2 * 31` output term assumes two P2WPKH outputs. Since #453 accepts Taproot destinations, a P2TR recipient output is 43 vB (undersized by 12) and a P2PKH recipient is 34 vB. The change output for a P2SH-P2WPKH sender is also P2SH (32 vB, not 31). The recipient output should be sized from its real scriptPubKey.

## Proposed fix

- Thread the resolved `addr_type` (already in scope at `bitcoin.py:654/663`) into `select_utxos` and size inputs per type: 68 (P2WPKH) / 91 (P2SH-P2WPKH) / 148 (P2PKH).
- Size the recipient output from its real scriptPubKey instead of a flat 31 (requires passing `to_address`/its script into `select_utxos`).
- Apply the same sizing to the duplicated logic in `scripts/test_embit_send.py`.

These can be split: the per-type **input** fix is small and unambiguously correct and can merge on its own; the **output** fix needs the destination script threaded in, so it's a slightly larger change.

## How to confirm

There is currently no assertion-based test pinning fee sizing to real vsize (`scripts/test_embit_send.py` is a manual harness that reproduces the same bug). Add a unit test that builds and signs a P2SH-P2WPKH-funded tx with embit (already a dependency), reads the real `tx.vsize`, asserts the current 68-based estimate is below it, then asserts the per-type estimate matches within rounding — turning the "91 not 68" claim into a green/red regression guard.

jaso0n0818:
## Problem

`select_utxos` used a flat 68 vB input size for all segwit types. P2SH-P2WPKH inputs are actually **91 vB** — the 22-byte redeemScript push in the non-witness scriptSig adds ~23 vB. With a 1.25× safety pad, multi-input P2SH-P2WPKH sends underestimate the fee starting at 4 inputs:

| Inputs (P2SH-P2WPKH) | est×1.25 | real | ok? |
|---|---|---|---|
| 1 | 176 | 164 | ✓ |
| 2 | 261 | 255 | ✓ |
| 3 | 346 | 346 | edge |
| 4+ | 431 | 437 | **✗** |

Under the override path (no 1.25× multiplier), *every* nested-segwit input is undersized by the full 23 vB regardless of count. Closes #459.

## Fix

****

Add a  class-level dict with per-type correct values:

Change  to  and derive vsize from the dict. The one call site () already had  in scope, so the boolean flag is removed.

****

Apply the same per-type map to the duplicated UTXO-selection logic (same bug, same fix).

## Tests

New  (8 tests):

- Asserts  constants (P2WPKH=68, P2SH-P2WPKH=91, P2PKH=148)
- Verifies single-input fees against the formula () × rate
- Confirms multi-input P2SH-P2WPKH fee > P2WPKH fee (main regression guard)
- Regression: asserts  (the pre-fix bug)

Closes #459

### DIFF
--- a/allways/chain_providers/bitcoin.py
+++ b/allways/chain_providers/bitcoin.py
@@ -701,10 +701,9 @@ class BitcoinProvider(ChainProvider):
                 bt.logging.info(f'Reusing prior tx {existing} from {my_address} → {to_address} ({amount} sat)')
                 return (existing, 0)
 
-            is_segwit = addr_type in ('p2wpkh', 'p2sh-p2wpkh')
             bt.logging.info(f'Sending from {addr_type} address: {my_address}')
 
-            coin_selection = self.select_utxos(utxos, amount, is_segwit, fee_rate_override=fee_rate_override)
+            coin_selection = self.select_utxos(utxos, amount, addr_type, fee_rate_override=fee_rate_override)
             if coin_selection is None:
                 return None
             selected, total_in, fee = coin_selection
@@ -826,10 +825,28 @@ class BitcoinProvider(ChainProvider):
         self._send_error('No UTXOs found for any address type')
         return None
 
-    def select_utxos(self, utxos, amount: int, is_segwit: bool, fee_rate_override: Optional[int] = None):
-        """Greedy UTXO selection. Returns (selected, total_in, fee) or None."""
+    # Input vsizes by address type (vbytes, from weight units):
+    #   P2WPKH:      41 B non-witness ×4 = 164  + 108 witness  = 272 wu → 68 vB
+    #   P2SH-P2WPKH: 64 B non-witness ×4 = 256  + 108 witness  = 364 wu → 91 vB
+    #   P2PKH:       ~148 B non-witness, no witness
+    _INPUT_VSIZE = {
+        ADDR_TYPE_P2WPKH: 68,
+        ADDR_TYPE_P2SH_P2WPKH: 91,
+        ADDR_TYPE_P2PKH: 148,
+        ADDR_TYPE_P2TR: 57,  # 41 B non-witness ×4 = 164 + 65 witness = 229 wu → 58 vB (rounded)
+    }
+    _INPUT_VSIZE_DEFAULT = 148  # conservative fallback for unknown types
+
+    def select_utxos(self, utxos, amount: int, addr_type: str, fee_rate_override: Optional[int] = None):
+        """Greedy UTXO selection sized per input address type.
+
+        Returns (selected, total_in, fee) or None on insufficient funds.
+        Uses per-type input vsizes so P2SH-P2WPKH inputs (91 vB) are no longer
+        undersized as P2WPKH (68 vB), which could cause below-target fee payment
+        on multi-input sends (issue #459).
+        """
         fee_rate = self.estimate_fee_rate(override=fee_rate_override)
-        input_vsize = 68 if is_segwit else 148
+        input_vsize = self._INPUT_VSIZE.get(addr_type, self._INPUT_VSIZE_DEFAULT)
         selected = []
         total_in = 0
         for utxo in sorted(utxos, key=lambda u: u['value'], reverse=True):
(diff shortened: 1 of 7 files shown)


jaso0n0818:
Rebased on latest `test`. Fixes P2SH-P2WPKH vsize underestimation (91 vB not 68 vB). Ready for review.

PR #465, opened 2026-06-09
fix(validator): serialize hotkey contract writes to prevent nonce collisions

jaso0n0818:
## Problem

The validator signs contract writes with the **same hotkey** over two separate substrate connections: the forward loop (`contract_client` on `self.subtensor`) and the axon handlers (`axon_contract_client` on `axon_subtensor`).

Both call `create_signed_extrinsic`, which auto-fetches the nonce via `AccountNonceApi.account_nonce` — the **best-block nonce**, which does not count pending pool txs. Within the same block window both clients fetch nonce `N`; one tx lands and the other is banned (**error 1012 Transaction is temporarily banned**), starving the forward loop and causing delivered swaps to be slashed. Closes #457.

### `allways/contract_client.py`

Add an optional `write_lock: threading.Lock` parameter to `AllwaysContractClient.__init__`. Inside `exec_contract_raw`, hold the lock across nonce-fetch + submit + inclusion. The pre-flight balance read is intentionally left **outside** the lock so reads remain parallel.

A `_NullContext` no-op is used when no write_lock is passed (backward-compatible).

### `neurons/validator.py`

Create one shared `self._write_lock = threading.Lock()` and pass it to both `contract_client` and `axon_contract_client` at construction. Lock order: `axon_lock → write_lock` (axon writes) and `write_lock → substrate_lock` (forward writes) — no cycle.

## Tests

New `tests/test_write_lock_serialization.py` (6 tests):

- Two clients sharing the same lock object store the same reference
- Client without `write_lock` completes `exec_contract_raw` normally  
- `write_lock` is held when `substrate_call(submit_extrinsic)` runs
- Balance read (`substrate_call` for account info) fires **before** the write_lock is acquired

Closes #457

### DIFF
--- a/allways/contract_client.py
+++ b/allways/contract_client.py
@@ -301,6 +301,16 @@ CONTRACT_ERROR_VARIANTS = {
 }
 
 
+class _NullContext:
+    """No-op context manager used when no write_lock is provided."""
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *_):
+        return False
+
+
 class ContractError(Exception):
     """Raised when a contract call fails.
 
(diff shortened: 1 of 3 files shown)


jaso0n0818:
Rebased on latest `test`. Serializes hotkey contract writes to prevent nonce collisions. Ready for review.

PR #480, opened 2026-06-15
fix(bitcoin): accept uppercase bech32 proof addresses

Yurii214:
## Summary
- Fixes #474 by normalizing uniform-case bech32/bech32m BTC addresses before proof address type detection.
- Keeps mixed-case bech32 rejected per BIP-173 while preserving existing base58 behavior.
- Applies the same canonicalization before mainnet re-encoding so uppercase segwit addresses verify through the BIP-137 proof path.

## Test plan
- `uv run pytest tests/test_bitcoin_signing.py`
- `uv run ruff check allways/chain_providers/bitcoin.py tests/test_bitcoin_signing.py`
- Direct repro check: uppercase `BC1Q...` returns `p2wpkh`; mixed-case `bc1Q...` returns `unknown`.

Fixes #474

PR #481, opened 2026-06-17
fix(bitcoin): resolve #470 nested-segwit scriptSig and #471 claim signer mismatch

### LINKED ISSUE #470: [Bug] P2SH-P2WPKH (nested-segwit) lightweight send assigns raw bytes to script_sig instead of a Script - final_tx.serialize() raises, so a nested-segwit miner can never broadcast a destination payout
## Summary

In the lightweight (`embit`) send path, the finalize step sets `final_tx.vin[i].script_sig` to `segwit_script.serialize()` — **raw `bytes`** — for the `p2sh-p2wpkh` case, while the legacy-P2PKH branch seven lines below correctly wraps its scriptSig in `Script(...)`. embit's `TransactionInput.script_sig` must be a `Script` (serialization calls `script_sig.serialize()`), so `final_tx.serialize()` raises `AttributeError: 'bytes' object has no attribute 'serialize'`, which is caught and the send returns `None`. A miner whose committed BTC address is **P2SH-P2WPKH** (`3…` mainnet / `2…` testnet) in `BTC_MODE=lightweight` therefore **cannot broadcast any destination payout** — every fulfillment fails, and the miner is slashed on every swap it takes.

## Affected code (branch `test`, `e360977`)

`allways/chain_providers/bitcoin.py:721-735` — the same field assigned two different types in adjacent branches of the same loop:
```python
for i, inp in enumerate(psbt.inputs):
    if is_segwit:
        for pub, sig in inp.partial_sigs.items():
            final_tx.vin[i].witness = Witness([sig, pub.sec()])
            if addr_type == 'p2sh-p2wpkh':
                final_tx.vin[i].script_sig = segwit_script.serialize()   # <-- raw bytes (BUG)
    else:
        from embit.script import Script
        for pub, sig in inp.partial_sigs.items():
            ...
            final_tx.vin[i].script_sig = Script(                          # <-- correctly a Script
                bytes([len(sig_bytes)]) + sig_bytes + bytes([len(pub_bytes)]) + pub_bytes
            )

raw_tx = final_tx.serialize().hex()   # :737 — calls each vin.script_sig.serialize()
```
P2SH-P2WPKH is a **supported** address type: `type_to_script` maps it (`bitcoin.py:647`), `is_segwit` includes it (`:663`), and the redeem-script is set on the PSBT input (`:693-695`). On the send-failure path, the `AttributeError` is swallowed and recorded as a generic send error (`return None`).

## Proof

`segwit_script.serialize()` returns `bytes`; `embit`'s `TransactionInput.script_sig` must be a `Script` object, and `Transaction.serialize()` calls `script_sig.serialize()` — `bytes` has no `.serialize()`, so the call raises. The legacy branch four lines down constructs `Script(<scriptSig bytes>)`, which is the correct type — the two branches cannot both be right for the same field. (Verified empirically against the repo's locked `embit`: wrapping as `Script(segwit_script.serialize())` serializes to a valid nested-segwit scriptSig `160014<20-byte-keyhash>` and a stable txid; the raw-`bytes` form raises.) Native P2WPKH and legacy P2PKH are unaffected — which is why it slipped through.

## Impact

100% broadcast failure for any nested-segwit miner running `BTC_MODE=lightweight`: it can never mark a swap fulfilled, so it is timeout-slashed on every swap it reserves — full, ongoing earnings loss for that miner. Missed because the existing tests cover BIP-137 message signing only, never transaction construction for this address type.

## Suggested fix

```python
final_tx.vin[i].script_sig = Script(segwit_script.serialize())
```
(import `Script` as the legacy branch already does). Add a serialization test for the nested-segwit finalize path (`addr_type == 'p2sh-p2wpkh'`) asserting `final_tx.serialize()` succeeds and produces the `160014…` scriptSig.

---

*Verified against `test` / `main` @ `e360977`. `git log -S 'script_sig = segwit_script.serialize()'` confirms this line is unchanged since the **initial commit** — never touched, despite this exact `finalize` block being actively maintained: PR #279 (merged) fixed an adjacent `embit` mutable-default leak here, #267 added the Esplora fallback, #225 added the traceback logging that now surfaces this very failure — none touched the `script_sig` type. Distinct from #459 (open — P2SH-P2WPKH **fee-sizing**, which assumes this send path works), #448/#460/#453 (address **validation** / Taproot / network), and #33/#34 (testnet P2SH **classification**). No open or merged PR/issue addresses the send-path `script_sig` type.*

### LINKED ISSUE #471: [Bug] `alw claim` signs `claim_slash` with the hotkey, but the contract gates it on `caller == swap.user` (the user's coldkey / receive address) — so a pending slash can't be claimed from the wallet that created the swap
## Summary

When a swap times out, the contract slashes the miner and pays the user directly; **if that direct transfer fails, the funds are parked in `pending_slashes` for the user to retrieve via `claim_slash`** (`lib.rs:1029-1043`). The contract's own doc comment states the intent — *"Claim a pending slash payout (user calls after failed transfer)"* (`lib.rs:1189`) — and enforces it with `caller == swap.user` (`lib.rs:1195`).

But `alw claim` submits `claim_slash` signed with **`wallet.hotkey`** (`contract_client.py:1286` → `_exec_logged` → `:1020 keypair=wallet.hotkey`), while `swap.user` is the user's **TAO address** — the **coldkey** for a TAO→BTC swap (`swap.py:948`) or the **receive address** for a BTC→TAO swap (`swap.py:933` / `axon_handlers.py:733-738`). A wallet's hotkey is neither. So for every recommended configuration `user != caller` holds and the claim returns `Error::InvalidStatus` (`lib.rs:1196`).

Result: a user running `alw claim` from the very wallet that created the swap is rejected by the contract's own access check, and the only shipped tool for the documented slash-recovery path cannot execute it. This is the same hotkey-vs-coldkey signing-identity mismatch already guarded on the TAO-send path (#270) and the collateral path (#380) — here on the slash-claim path.

## Affected code (branch `test`, `e360977`; identical on `main`, `db9ab0d`)

**The CLI signs with the hotkey** — `allways/contract_client.py:1286` → `_exec_logged` (`:1011-1022`):
```python
def claim_slash(self, wallet: bt.Wallet, swap_id: int) -> str:
    return self._exec_logged('claim_slash', wallet, f'Slash claimed for swap {swap_id}', {'swap_id': swap_id})
# _exec_logged:
tx_hash = self.exec_contract_raw(method, args=args, keypair=wallet.hotkey, value=value)   # :1020
```
Every `contract_client` write signs with `wallet.hotkey` (`grep keypair=wallet.coldkey` → 0 matches); there is no coldkey-signed path.

**The contract requires `caller == swap.user`** — `smart-contracts/ink/lib.rs:1189-1197`:
```rust
/// Claim a pending slash payout (user calls after failed transfer)
pub fn claim_slash(&mut self, swap_id: u64) -> Result<(), Error> {
    let caller = self.env().caller();
    let (user, amount) = self.pending_slashes.get(swap_id).ok_or(Error::NoPendingSlash)?;
    if user != caller {
        return Err(Error::InvalidStatus);
    }
    ...
```
`pending_slashes` is keyed to `swap.user`, populated only when the direct timeout payout fails — `lib.rs:1029-1038`:
```rust
let actual_slash = this.apply_collateral_penalty(swap.miner, swap.tao_amount);
if actual_slash > 0 {
    if this.env().transfer(swap.user, actual_slash).is_ok() {
        ...                                                          // direct payout to swap.user
    } else {
        this.pending_slashes.insert(swap_id, &(swap.user, actual_slash));   // <-- claim path
```

**`swap.user` is the user's TAO address, never the hotkey** — `allways/validator/axon_handlers.py:733-738` derives it (`to_address` for BTC→TAO, `from_address` for TAO→BTC) and stores it on-chain as `swap.user`; the CLI sets those addresses in `allways/cli/swap_commands/swap.py`:
- **TAO→BTC** source (`:948`): `user_from_address = wallet.coldkeypub.ss58_address` → `swap.user` = the **coldkey**.
- **BTC→TAO** receive (`:933`): `click.prompt('Your TAO receive address')` → `swap.user` = the **user-supplied receive address**.

**The display reinforces the wrong identity** — `allways/cli/swap_commands/claim.py:65-66`:
```python
console.print(f'  Claiming:   {wallet.hotkey.ss58_address}')
console.print('[dim]  Only the original swap user can claim; others will be rejected on-chain.[/dim]')
```
It prints the hotkey as the claimant and asserts the user can claim — then signs with the hotkey, which is not `swap.user`.

## Why it fails (proof)

- `caller` = the extrinsic signer = `wallet.hotkey` (the only keypair `contract_client` ever signs with).
- `swap.user` ∈ { user coldkey (TAO→BTC), user receive address (BTC→TAO) }.
- A wallet's hotkey and coldkey are distinct accounts, and a separately-supplied receive address is a third — so `user != caller` and the claim is rejected with `InvalidStatus` deterministically.

**Honest scope (the one exception):** the claim succeeds only if `swap.user` happens to equal `wallet.hotkey.ss58_address` — i.e. a BTC→TAO user who typed *their own hotkey* as the receive address. That configuration is explicitly discouraged by the existing guard in #270 (because TAO transfers sign from the coldkey, not the hotkey), so the claim is broken precisely for the *recommended* setups — the self-custody coldkey source and any distinct receive address.

## Reachability / severity

`pending_slashes` is populated only when the **direct** timeout transfer to `swap.user` fails (`lib.rs:1031` `else`). The realistic trigger is a **BTC→TAO** swap whose `swap.user` is a fresh TAO receive address below the existential deposit, so `env().transfer` cannot create it and the payout falls through to `pending_slashes` — which is exactly the make-whole path that exists for "the user's payout couldn't be delivered." When any pending slash exists, `alw claim` is the only shipped recovery tool, and it is signed with the wrong key. So this is not cosmetic: the documented recovery path is unusable from shipped tooling for the recommended configurations, with the user's make-whole funds (up to the slashed `tao_amount`) left in `pending_slashes`. (For the TAO→BTC case the direct payout to the funded coldkey usually succeeds, so a pending slash is rarer there — but when one exists, the fix below makes it claimable.)

## Relationship to prior work (distinct, and the precedents point the same way)

- **PR #27** ("Fix permanent fund lock in pending slashes", closed) targeted a *different* failure point: a **reaped** user account makes the `env().transfer(caller, amount)` **inside** `claim_slash` (`lib.rs:1200`) fail — i.e. it **passes** the `caller == user` gate, then the transfer fails. It was closed on the argument that the lock isn't permanent because *"the user holds the key"* and can re-fund their account and claim. That rebuttal **assumes the user can call `claim_slash` from their own (`swap.user`) account** — which the shipped `alw claim` does not allow, because it signs with the hotkey. This issue is that earlier, orthogonal gate (`:1195`); it is the precondition #27's resolution relies on, and its fix is **client-side** (no contract change, sidestepping #27's owner-privilege objection).
- **PR #163** (merged) reworked this command's *messaging* ("only the original user can claim") without touching the signing key — it added the copy that masks this mismatch.
- **PR #270** (merged) and **PR #380** (merged) both fix the **same class** of bug — hotkey vs coldkey signing identity — on the TAO-send and collateral paths respectively, **client-side** (a warning / a new command). This is that mismatch on the claim path.

Exhaustive dedup (issues + PRs, all states, GraphQL): no issue or PR addresses the `claim_slash` signing identity. The `claim_slash`/`pending_slashes` hits are #27 (above) and `_exec_logged` refactors (#91/#94/#104/#89); none change which key signs the claim.

## Suggested fix

`claim_slash` is a **user** action; its on-chain caller must equal `swap.user`:
1. **Sign the claim with the key that matches `swap.user`.** For the self-custody TAO→BTC case (`swap.user` = coldkey), sign `claim_slash` with `wallet.coldkey` (a coldkey-signed path for this user-facing call), consistent with the "TAO actions sign from the coldkey" model behind #270/#380.
2. **Pre-flight the BTC→TAO case.** Read the swap, derive `swap.user`; if it matches a key the wallet controls, sign with it; otherwise refuse with a clear message — *"this slash is claimable only from `<swap.user>`; submit the claim from that account"* — instead of broadcasting a guaranteed-to-fail hotkey transaction.
3. Fix `claim.py:65` to display `swap.user` (the required claimant), not `wallet.hotkey`.

## Suggested test (`tests/`)

- TAO→BTC: a swap whose `swap.user` is the wallet coldkey ss58 with a populated `pending_slashes`; assert `alw claim` signs `claim_slash` with the **coldkey** (`caller == swap.user`) and the contract accepts.
- BTC→TAO with an address the wallet does not control: assert the command refuses pre-flight ("claim from `<addr>`") and does **not** broadcast.
- Regression guard: assert `claim_slash` is never signed with `wallet.hotkey`.

---

*Verified against `test` @ `e360977` and `main` @ `db9ab0d` (identical). `claim_slash` has signed with `wallet.hotkey` since the initial commit; the contract gate `caller == swap.user` (`lib.rs:1195`) and `pending_slashes` keying (`:1038`) are current. PR #163 (merged) changed only this command's messaging; PR #27 (closed) targeted the reaped-account transfer **inside** `claim_slash`, not the client signing identity; PR #270 / #380 (merged) fix the same hotkey-vs-coldkey signing mismatch on other paths, client-side. GraphQL dedup across all issues/PRs shows nothing touching the claim signing key. Fix is Python-only (`claim.py` + a coldkey-signed `claim_slash` path).*

Yurii214:
## Summary
- Fixes #470 by setting nested-segwit (`p2sh-p2wpkh`) `script_sig` as `Script(...)` in lightweight BTC send finalization so serialization/broadcast succeeds.
- Fixes #471 by allowing `claim_slash` to use an explicit signer and selecting the signer in CLI claim flow based on `swap.user` (`coldkey` or `hotkey`).
- Adds a claim preflight guard that exits early with a clear message when the current wallet cannot sign for the swap user account.

## Tests
- Added nested-segwit send regression coverage in `tests/test_bitcoin_signing.py`.
- Added explicit signer/default signer tests in `tests/test_contract_client.py`.
- Added CLI claim signer-selection tests in new `tests/test_claim_command.py`.

## Validation run
- `uv run pytest tests/test_bitcoin_signing.py tests/test_contract_client.py tests/test_claim_command.py`
- `uv run ruff check allways/chain_providers/bitcoin.py allways/contract_client.py allways/cli/swap_commands/claim.py tests/test_bitcoin_signing.py tests/test_contract_client.py tests/test_claim_command.py`

Fixes #470
Fixes #471

### DIFF
--- a/allways/chain_providers/bitcoin.py
+++ b/allways/chain_providers/bitcoin.py
@@ -720,10 +720,12 @@ class BitcoinProvider(ChainProvider):
             final_tx = psbt.tx
             for i, inp in enumerate(psbt.inputs):
                 if is_segwit:
+                    from embit.script import Script
+
                     for pub, sig in inp.partial_sigs.items():
                         final_tx.vin[i].witness = Witness([sig, pub.sec()])
                         if addr_type == 'p2sh-p2wpkh':
-                            final_tx.vin[i].script_sig = segwit_script.serialize()
+                            final_tx.vin[i].script_sig = Script(segwit_script.serialize())
                 else:
                     from embit.script import Script
 
--- a/allways/cli/swap_commands/claim.py
+++ b/allways/cli/swap_commands/claim.py
@@ -38,12 +38,12 @@ def claim_command(swap_id: int, yes: bool):
         print_contract_error('Failed to read pending slash', e)
         return
 
-    if pending_rao == 0:
-        try:
-            swap = client.get_swap(swap_id)
-        except ContractError:
-            swap = None
+    try:
+        swap = client.get_swap(swap_id)
+    except ContractError:
+        swap = None
 
+    if pending_rao == 0:
         if swap is not None and swap.status in (SwapStatus.ACTIVE, SwapStatus.FULFILLED):
             console.print(
                 f'[yellow]Swap #{swap_id} is still in progress (status: {swap.status.name}).[/yellow]\n'
@@ -60,9 +60,30 @@ def claim_command(swap_id: int, yes: bool):
         )
         return
 
+    signer_keypair = wallet.hotkey
+    signer_ss58 = wallet.hotkey.ss58_address
+    required_claimant = None
+    if swap is not None:
+        required_claimant = swap.user_hotkey
+        if required_claimant == wallet.coldkeypub.ss58_address:
+            signer_keypair = wallet.coldkey
+            signer_ss58 = wallet.coldkeypub.ss58_address
+        elif required_claimant == wallet.hotkey.ss58_address:
+            signer_keypair = wallet.hotkey
+            signer_ss58 = wallet.hotkey.ss58_address
+        else:
+            console.print(f'  Swap user:  [yellow]{required_claimant}[/yellow]')
+            console.print(
+                '[red]This wallet cannot claim this slash.[/red]\n'
+                '[dim]Claim must be signed by the original swap user account above.[/dim]\n'
+            )
+            return
+
     console.print(f'  Swap ID:    {swap_id}')
     console.print(f'  Amount:     [green]{from_rao(pending_rao):.4f} TAO[/green] ({pending_rao} rao)')
-    console.print(f'  Claiming:   {wallet.hotkey.ss58_address}')
+    if required_claimant:
+        console.print(f'  Swap user:  {required_claimant}')
+    console.print(f'  Claiming:   {signer_ss58}')
     console.print('[dim]  Only the original swap user can claim; others will be rejected on-chain.[/dim]\n')
 
     if not yes and not click.confirm('Confirm claiming slash?'):
@@ -71,7 +92,7 @@ def claim_command(swap_id: int, yes: bool):
 
     try:
         with loading('Submitting transaction...'):
-            client.claim_slash(wallet=wallet, swap_id=swap_id)
+            client.claim_slash(wallet=wallet, swap_id=swap_id, keypair=signer_keypair)
         console.print(f'[green]Successfully claimed {from_rao(pending_rao):.4f} TAO from swap #{swap_id}![/green]\n')
     except ContractError as e:
         print_contract_error('Failed to claim slash', e)
--- a/allways/contract_client.py
+++ b/allways/contract_client.py
@@ -1015,9 +1015,10 @@ class AllwaysContractClient:
         log_msg: str,
         args: Optional[dict] = None,
         value: int = 0,
+        keypair: Optional[Any] = None,
     ) -> str:
         self.ensure_initialized()
-        tx_hash = self.exec_contract_raw(method, args=args, keypair=wallet.hotkey, value=value)
+        tx_hash = self.exec_contract_raw(method, args=args, keypair=keypair or wallet.hotkey, value=value)
         bt.logging.info(f'{log_msg}: {tx_hash}')
         return tx_hash
 
@@ -1283,8 +1284,14 @@ class AllwaysContractClient:
         """Deactivate a miner directly on contract (permissionless)."""
         return self._exec_logged('deactivate', wallet, f'Miner {miner} deactivated', {'miner': miner})
 
-    def claim_slash(self, wallet: bt.Wallet, swap_id: int) -> str:
-        return self._exec_logged('claim_slash', wallet, f'Slash claimed for swap {swap_id}', {'swap_id': swap_id})
+    def claim_slash(self, wallet: bt.Wallet, swap_id: int, keypair: Optional[Any] = None) -> str:
+        return self._exec_logged(
+            'claim_slash',
+            wallet,
+            f'Slash claimed for swap {swap_id}',
+            {'swap_id': swap_id},
+            keypair=keypair,
+        )
 
     # =========================================================================
     # Admin Transaction Functions (Write — owner-only)
--- a/tests/test_bitcoin_signing.py
+++ b/tests/test_bitcoin_signing.py
@@ -273,6 +273,47 @@ class TestBroadcastedTxidsTracking:
         provider = make_lightweight_provider()
         assert provider.broadcasted_txids == set()
 
+    def test_nested_segwit_lightweight_send_serializes_scriptsig(self):
+        """Regression guard for #470: nested-segwit finalize must store a Script
+        object on script_sig, not raw bytes, so tx serialization succeeds."""
+        provider = make_lightweight_provider()
+        from_address, _, _ = sign_message(TEST_WIF, 'p2wpkh-p2sh', 'x', deterministic=True)
+        to_address, _, _ = sign_message(TEST_WIF, 'p2wpkh', 'y', deterministic=True)
+
+        utxo_resp = MagicMock()
+        utxo_resp.raise_for_status.return_value = None
+        utxo_resp.json.return_value = [{'txid': '11' * 32, 'vout': 0, 'value': 80_000}]
+
+        captured = {}
+
+        def fake_post(path: str, data: str, timeout: int = 30):
+            captured['raw_tx'] = data
+            return SimpleNamespace(status_code=200, text='aa' * 32)
+
+        with (
+            patch.dict(os.environ, {'BTC_PRIVATE_KEY': TEST_WIF}, clear=False),
+            patch.object(provider, 'btc_api_get', return_value=utxo_resp),
+            patch.object(provider, 'btc_api_post', side_effect=fake_post),
+        ):
+            result = provider.send_amount_lightweight(
+                to_address=to_address,
+                amount=10_000,
+                from_address=from_address,
+                fee_rate_override=1,
+            )
+
+        assert result == ('aa' * 32, 0)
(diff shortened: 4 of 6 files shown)


Yurii214:
Hi — gentle bump on this PR. Like #480, the workflows here are still waiting on maintainer approval, so checks haven’t run yet.

This bundles two related bugfixes:

- **#470** — P2SH-P2WPKH lightweight sends now set `script_sig` as `Script(...)` instead of raw bytes, fixing serialization/broadcast failures.
- **#471** — `claim_slash` and the CLI `claim` flow now sign with the account that matches `swap.user` (coldkey or hotkey), with an early exit if the wallet can’t sign.

Tests added/updated:
- `tests/test_bitcoin_signing.py` (nested-segwit send regression)
- `tests/test_contract_client.py` (explicit vs default signer for `claim_slash`)
- `tests/test_claim_command.py` (CLI signer selection + refusal path)

Locally:
- `uv run pytest tests/test_bitcoin_signing.py tests/test_contract_client.py tests/test_claim_command.py`
- `uv run ruff check` on the touched files (all green)

Could you approve the workflow run when you have time? Happy to split this into two PRs if that’s easier to review. Thanks!

PR #488, opened 2026-06-22
feat(solana): charge churn fee on remove_quote (close-recreate bypass)

LandynDev:
Closes a bypass in the #484 anti-flashing quote fee.

## The hole
`set_quote` detects "first creation" (free) by a zeroed quote PDA. But `remove_quote` does `close = miner` — it deletes the PDA and refunds rent. So:

1. `remove_quote` → PDA closed, rent back
2. `set_quote` → fresh PDA → `miner == default` → **fee = 0**

Net cost to flash-bypass ≈ tx fees + a rent round-trip (a wash). The 0.01 SOL churn fee is fully dodged.

## Fix
`remove_quote` charges the **same** `quote_update_fee`, keyed off how long the quote stood. Disturbing a fresh quote costs the same whether you overwrite it in place or close-then-recreate; a long-standing quote still removes free.

- `RemoveQuote` gains `vault` + `system_program` (mirrors `SetQuote`)
- `QuoteRemoved` gains `remove_fee`

## Tests
- New `test_remove_fresh_quote_charges_churn_fee`: fresh remove → tier-1; stale remove → free
- `test_remove_quote_closes_and_refunds`: warps past the decay window so it stays a pure rent-refund check
- On-chain + integration remove-quote helpers updated for the new accounts

> Built against `contract-v2` (host `cargo check` clean). LiteSVM/e2e need the SBF toolchain — not run locally.

PR #490, opened 2026-06-22
feat(solana): PR-484 review fixes + Treasury/Vault separation

anderdc:
Follow-up to #484 (merged): the review fixes that didn't make the squash, plus a Treasury/Vault separation and the consolidation of the closed #488. Branched off `contract-v2` and ingests #489 (comment compression).

## Correctness / fund-safety
- **Over-collateralization gated at pool entry** (`open_or_request`): a miner must hold `1.10x` the swap size up front. Collateral can't drop while busy, so `vote_initiate` can no longer strand a user whose source funds are already committed. (484 review #1)
- **`busy ⟹ active` invariant:** `vote_deactivate` now refuses a busy miner (mirrors self-`deactivate`), so a miner mid-pool/reservation/swap can't be deactivated. Nothing clears `busy_until` early, so the invariant holds cleanly.
- **No admin cancel ops.** Evaluated restoring `cancel_pool`/`cancel_reservation` (deleted in #485) but **dropped them**: no permanent stuck state exists — `resolve_pool` is permissionless (an open pool always progresses) and a reservation self-expires at `reservation_ttl`. They were only early-clear accelerators and the sole paths that cleared `busy_until` manually (the footgun behind a fund-safety bug caught in review). `resolve_pool` needs no active check (documented invariant); `vote_initiate` keeps a defensive `active` check as the funds-commit backstop.
- **Contradictory config bounds rejected** (`set_min/max_swap_amount`, collateral) via shared `validate.rs`, used by **both** `initialize` and the setters so the two write paths can't diverge. (#6, #8)

## Treasury / Vault separation
- Collateral and subnet revenue no longer share an account. **`Vault`** holds only collateral (trustless — leaves only via the owning miner's withdraw or a slash to the wronged user); new **`Treasury` PDA** holds only revenue (1% confirm fee, reservation fee, quote churn fee). Timeout slash still pays the **user** from the Vault. Split invariants: `vault.lamports == rent + total_collateral`, `treasury.lamports == rent + total`.
- Also removes the `set_quote` write-lock on the collateral vault (a parallelism win).

## Anti-flash quote fee (supersedes #488)
- Creation is **free**; updates pay the decaying churn fee; `remove_quote` charges the same decaying fee → Treasury, closing the remove+recreate dodge without taxing first-time quotes. (Adopted @LandynDev's remove-side mechanism from #488, routed to the new Treasury.) Removal can cost the fee by design — posting a quote affirms a commitment to honor it.

## Other
- `tunables.rs` folded into `constants.rs` (one home for economic levers); `DEFAULT_FULFILLMENT_TIMEOUT_SECS = 14400` (4h) canonical deploy default.

## Tests
- LiteSVM **65 passed / 17 ignored / 0 failed**; `e2e.sh` **24/24** against a live validator (build + deploy + on-chain).
- Coverage: entry over-collateral gate, busy-lock deactivation, bounds validation, quote create-free/remove-fee, treasury lamport conservation.


LandynDev (approved):
Reviewed for correctness: Vault/Treasury split holds both invariants on every fund path; busy⟹active invariant sound; shared validate.rs wired into initialize + all setters. Approving. Follow-up: bump CONFIG_VERSION; fix stale vote_initiate comment.

PR #491, opened 2026-06-22
refactor(solana): per-miner collateral vaults

anderdc:
**Stacked on #490** (`solana-review-followups`). Until #490 merges, this PR's diff shows #490's changes too; once #490 lands in `contract-v2`, it collapses to just the per-miner-collateral change. Review #490 first.

## What

Replace the single shared collateral `Vault` with **per-miner `CollateralVault` PDAs** (`seeds=[b"collateral", miner]`). Each miner's collateral lamports live in their own account; `MinerState.collateral` stays the authoritative ledger. The `Treasury` PDA (subnet revenue) is unchanged.

## Why

The shared `Vault` was both a write-serialization point and a cheap-to-grief congestion target: a miner spamming self-deposits could eat the vault's ~12M-CU/block budget and **delay swap settlement** (`confirm`/`timeout` wrote the same account). Per-miner vaults:
- **isolate the DOS blast radius** — a miner's deposit/withdraw spam only congests their own PDA;
- **parallelize** collateral ops across different miners;
- make custody **literal** — a miner's collateral is in their own account, movable only by their own `withdraw_collateral` or a quorum slash.

No fund-access change: deposit/withdraw were already signer-scoped to the owner.

## Changes
- New `CollateralVault { bump }`; `Vault` struct + `VAULT_SEED` removed.
- `post_collateral`/`withdraw_collateral` move lamports between the miner and *their own* vault (lazily created on first deposit, miner-paid rent).
- `confirm_swap`: 1% fee → Treasury from the miner's vault; `timeout_swap`: slash → user from the miner's vault.
- `apply_penalty` drops the shared-total bookkeeping (per-account invariant: `lamports == rent + collateral`).
- `initialize` no longer creates a shared vault.

## Tests
- LiteSVM **65/65**; `e2e.sh` **24/24** (build + deploy + on-chain, exercising the per-miner vaults end-to-end).
- Per-miner invariant assertions (`collateral_vault(miner).lamports == rent + MinerState.collateral`); removed the obsolete shared-`total_collateral` checks.

### DIFF
--- a/smart-contracts/solana/programs/allways_swap_manager/src/constants.rs
+++ b/smart-contracts/solana/programs/allways_swap_manager/src/constants.rs
@@ -4,9 +4,10 @@ use anchor_lang::prelude::*;
 #[constant]
 pub const CONFIG_SEED: &[u8] = b"config";
 
-/// PDA seed for the singleton native-SOL collateral vault (`seeds = [VAULT_SEED]`).
+/// PDA seed prefix for a per-miner native-SOL collateral vault (`seeds = [COLLATERAL_SEED, miner]`) —
+/// each miner's collateral in its own account: trustless custody + no shared-vault contention.
 #[constant]
-pub const VAULT_SEED: &[u8] = b"vault";
+pub const COLLATERAL_SEED: &[u8] = b"collateral";
 
 /// PDA seed prefix for per-miner state (`seeds = [MINER_SEED, miner_pubkey]`).
 #[constant]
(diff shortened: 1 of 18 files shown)


LandynDev (approved):
Per-miner collateral vaults: fund-safe. collateral_vault/miner_state/swap all derive from the same miner key + swap has_one=miner (fee/slash unsubstitutable); timeout payout bound via swap.user==user.key(); debits clamped so a vault never drops below rent. Approving.

PR #492, opened 2026-06-22
feat(solana): dynamic pool bids (in-window request updates)

anderdc:
**Stacked on #491 → #490.** Until those merge, this PR's diff includes their changes too; once they land in `contract-v2` it collapses to just this change. Review order: #490 → #491 → this.

## What

A validator gets **one shot** per reservation pool today — a second `open_or_request` from the same validator while the pool is open is rejected with `AlreadyRequested`. This lets a validator **update its bid in place** while the window is still open (upsert), frozen once the pool resolves.

## Why

The miner's rate is **pinned at pool open**, so the intended workflow is for a validator to run its user-selection *after* a pool opens (bidding against a now-stable rate) and refine as better users arrive. Part B of the reservation/initiation evolution.

- **No lottery gaming:** draw odds are stake-weighted, not amount/time-weighted.
- **Miner protected:** terms stay pinned regardless; an updated bid re-runs the amount-bounds + over-collateral checks (top of the handler).
- **`busy_until` unchanged:** set only at OPEN, never extended by an update.

## Changes (one handler)

`open_or_request.rs`:
- **JOIN branch** upserts: a repeat call from the same validator replaces its `Request` in place; a new validator is appended (subject to `MAX_VALIDATORS`).
- **Fee** is now charged only on a *fresh* entry (open or first join); a same-validator in-window **update is free** (refinement isn't taxed). Different validators each still pay.
- `ErrorCode::AlreadyRequested` left defined (now unused) to avoid IDL churn.

## Tests
- `test_validator_can_update_request` (replaces the old dup-reject test), `test_update_reflected_in_reservation`, `test_update_is_free`, `test_update_after_window_close_fails`.
- LiteSVM **68/68**; `e2e.sh` **24/24** (build + deploy + on-chain).

### DIFF
--- a/smart-contracts/solana/programs/allways_swap_manager/src/instructions/open_or_request.rs
+++ b/smart-contracts/solana/programs/allways_swap_manager/src/instructions/open_or_request.rs
@@ -129,24 +129,35 @@ pub fn handler(
     let active_reservation = resv.reserved_until != 0 && resv.reserved_until >= now;
     require!(!active_reservation, ErrorCode::MinerReserved);
 
-    // Flat, non-refundable anti-spam fee: validator → treasury (subnet revenue, every call).
-    let fee = ctx.accounts.config.reservation_fee_lamports;
-    transfer(
-        CpiContext::new(
-            ctx.accounts.system_program.key(),
-            Transfer {
-                from: ctx.accounts.validator.to_account_info(),
-                to: ctx.accounts.treasury.to_account_info(),
-            },
-        ),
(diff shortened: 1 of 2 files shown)


LandynDev (approved):
Dynamic pool bids (upsert): correct. Bounds + 1.1x over-collateral re-run on every update; draw is stake-weighted so updates can't game odds; OPEN clears requests (no stale-fee dodge); clean freeze at closes_at. Approving.

PR #493, opened 2026-06-23
feat(solana): deadline extensions (reservation + fulfillment)

LandynDev:
## What

Lets any validator slide a reservation's `reserved_until` or a swap's `timeout_at` **forward** while it waits on slow chain confirmation — so a correct miner isn't false-slashed when a BTC confirmation lands after a tight deadline.

Two new instructions: `extend_reservation`, `extend_timeout`. **No quorum.**

## Why it's safe without quorum

An extension moves no funds. Money only moves at `confirm_swap` / `timeout_swap`, which stay quorum-gated. So a rogue validator can at most **delay** a slash up to a fixed ceiling — never cause one, never block one past the ceiling.

Two guards carry it:
- **monotonic** — target must be later than the current deadline (can't shorten a miner's runway)
- **ceiling** — `max_extend_at`, frozen at creation = deadline + budget (can't hold a miner hostage)

No count cap / no per-extension cap: the total ceiling bounds the worst case on its own. Validators nudge adaptively as confirmations land.

## Config
- `max_total_extension_secs` — default **90 min**, admin-tunable within **[30 min, 120 min]**.
- Reset base deadlines to the intended **10 min** each side (`DEFAULT_FULFILLMENT_TIMEOUT_SECS` 4h→10m, new `DEFAULT_RESERVATION_TTL_SECS` 10m), mirroring ink!'s 50-block default. Model: tight base, extend as needed → ≤ ~100 min.
- Extensions ignore `halted` (in-flight swaps must finish).

## Tests
`test_extensions.rs`: happy path, ceiling cap, monotonic reject, validator-only, Active + Fulfilled. Bounds unit test passes; full integration run needs `build-sbf` in the dev env.

## Follow-ups (not in this PR)
Python `contract_client` methods, `optimistic_extensions.py` collapse to a single `extend` call, `event_watcher` for the 2 new events. Deployer/dev-env must pass the 10m init args.

### DIFF
--- a/smart-contracts/solana/programs/allways_swap_manager/src/consensus.rs
+++ b/smart-contracts/solana/programs/allways_swap_manager/src/consensus.rs
@@ -57,6 +57,16 @@ pub fn weights_hash(validators: &[ValidatorInfo], weights: &[u64]) -> [u8; 32] {
     hashv(&refs).to_bytes()
 }
 
+/// Assert `validator` is whitelisted — for consensus-free validator actions (deadline extensions)
+/// that gate on membership but open no vote round.
+pub fn ensure_validator(config: &Config, validator: &Pubkey) -> Result<()> {
+    require!(
+        config.validators.iter().any(|v| &v.key == validator),
+        ErrorCode::NotValidator
+    );
+    Ok(())
+}
+
 /// Record a validator's vote into `round`; returns true iff quorum is now reached.
 /// (Re)initializes a fresh or stale round, binds params via `bound_hash`, and dedupes voters.
 pub fn record_vote<'info>(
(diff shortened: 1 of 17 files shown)


anderdc:
Ran the suite against this branch — `test_extensions` is 8/8 green, the feature works. Two pre-existing version assertions need updating for the `CONFIG_VERSION` 6→7 bump, though, or the build fails:

- `programs/allways_swap_manager/tests/test_initialize.rs:60` — `assert_eq!(config.version, 6)` → `7` (LiteSVM unit test, fails)
- `programs/allways_swap_manager/tests/integration_onchain.rs:510` — `assert_eq!(cfg.version, 6, "schema version")` → `7` (on-chain test, fails in e2e)

Both are just the stale asserts from the schema bump, not logic. Flip them to `7` and the suite goes green.

LandynDev:
Pushed `adc3782` — flipped both stale version asserts to 7 (`test_initialize.rs:60`, `integration_onchain.rs:510`). Thanks for running the suite; glad `test_extensions` is 8/8 green.

PR #494, opened 2026-06-24
feat(solana): scoring read-surface + freshness source-replay (A1–A5) for the off-chain refactor

anderdc:
## Summary

Adds the **on-chain read-surface** the off-chain validator refactor will build against — making the validator dumber/more trustless by moving derivable state onto the contract — plus replaces the permanent source-replay marker with a freshness bound, and hardens the claim path. Additive over `contract-v2` (no behavior change to existing flows). `CONFIG_VERSION` → 10.

## What's in it

| Task | Change |
|---|---|
| **A1** | `MinerState.successful_swaps`/`failed_swaps` counters — the flat eligibility gate reads these (no off-chain credibility ledger). |
| **A2** | `MinerDirectionStats` PDA + richer `SwapCompleted` — per-direction realized volume/VWAP for the "quality volume" scoring half. |
| **A3** | `submit_swap_claim` + `PendingAttestation` — the source-tx claim moves on-chain (deletes the brittle off-chain `pending_confirms` queue). Payout is **pinned** in the reservation at lottery time (front-run can't redirect it). `close_stale_claim` reaps orphans. |
| **A4** | **Freshness source-replay** — removed the permanent `TxMarker` (unbounded forever-rent); added `Reservation.created_at` as the source-freshness bound (dest bound = existing `swap.initiated_at`). The actual checks are validator-side (see below). |
| **A5** | `bind_hotkey` + `Binding` PDA — mutual-signature hotkey↔Solana-pubkey identity (strikes attribute to the right UID; rotation can't dodge them). |
| **fix** | `submit_swap_claim` gated to whitelisted validators (relay) — kills the front-run *squat* of the one-claim slot + junk-claim RPC-amplification, mirroring the proven ink! "only validators put a tx on-chain" model. Participation stays open. |
| **B** | Pool entry confirmed **already permissionless** (no validator gate; non-validator routers get lottery weight 0 → uniform/lose to validators). Renamed the misleading `Request.validator` → `router` for clarity; no behavior change. |
| **chore** | Pre-PR cleanup: dropped dead `Reservation.bound_hash`; fixed stale comments. |

## ⚠️ Load-bearing follow-up (validator / Phase B, before mainnet)

A4 removed the contract-level source-replay guard. The replacement — a validator **freshness check** (`from_tx` block-time > `Reservation.created_at`; `to_tx` block-time > `swap.initiated_at`) — lives off-chain and **must be implemented + tested before mainnet** (tracked in `SOLANA_VALIDATOR_CHANGES.md`). Safe pre-launch/devnet meanwhile.

## Deferred (own design pass)

A cheap **miner-DoS**: a single ~0.02-SOL junk `open_or_request` busies a miner for the pool window + reservation TTL (the junk winner can't claim → miner idle until expiry); repeatable. Pre-existing (entry was already open), not introduced here. Mitigation = fee-sizing / requester bond / only-the-winner-busies — revisit during Phase B.

## Testing

- **84 LiteSVM unit tests** green (`cargo test -p allways_swap_manager`).
- **25/25 on-chain integration tests** green (`e2e.sh` harness, solana-test-validator).


smart-contracts/solana/programs/allways_swap_manager/src/state.rs
```diff
@@ -246,12 +260,58 @@ pub struct MinerQuote {
...
+pub struct MinerDirectionStats {
+    pub miner: Pubkey,
+    #[max_len(MAX_CHAIN_LEN)]
+    pub from_chain: String,
+    #[max_len(MAX_CHAIN_LEN)]
+    pub to_chain: String,
+    /// Count of completed (confirmed) swaps in this direction.
+    pub completed: u32,
+    /// Sum of the collateral-backed SOL size over completed swaps (lamports).
+    pub total_sol_amount: u128,
```
LandynDev: **A2 — `total_sol_amount` may be redundant here (not a blocker).**

The realized signal the scoring layer needs is the per-direction `total_from_amount` / `total_to_amount`. Because this PDA is keyed by `(miner, from_chain, to_chain)`, those two are already **asset-pure** — `total_from_amount` is always in `from_chain`’s unit, `total_to_amount` always in `to_chain`’s — so the realized VWAP works without any normalization.

`total_sol_amount` is the only field needing a common unit, and that normalization happens **off-chain** in the scoring validator, which already runs a price feed for the rate-vs-market benchmark. It can re-derive SOL-notional from `total_from_amount × price` for free. Two downsides to keeping it on-chain: (a) it duplicates something the validator computes anyway, and (b) it stops being a common unit the moment collateral goes multi-asset — a TAO-collateralized miner’s `sol_amount` isn’t comparable to a SOL-collateralized one’s, so summing across miners mixes units. Since this is permanent state and the `sol_*` name bakes in the collateral=SOL assumption that split-collateral will unwind, dropping it (and letting from/to carry the realized record) keeps the PDA asset-agnostic for free. The at-time notional, if ever wanted, is already in the extended `SwapCompleted` event. Worth settling before this PDA shape locks.

smart-contracts/solana/programs/allways_swap_manager/src/instructions/confirm_swap.rs
```diff
@@ -101,6 +130,32 @@ pub fn handler(ctx: Context<ConfirmSwap>, swap_key: [u8; 32]) -> Result<()> {
...
+        let stats_bump = ctx.bumps.direction_stats;
+        let stats = &mut ctx.accounts.direction_stats;
+        if stats.miner == Pubkey::default() {
+            stats.miner = miner;
+            stats.from_chain = from_chain.clone();
+            stats.to_chain = to_chain.clone();
+            stats.bump = stats_bump;
+        }
+        stats.completed = stats.completed.saturating_add(1);
+        stats.total_sol_amount = stats
```
LandynDev: Tied to the `MinerDirectionStats` note in state.rs: this accrual is the only writer of `total_sol_amount`. If that field is dropped as asset-agnostic cleanup, this block goes with it and the `from`/`to` accruals below carry the realized record unchanged.

smart-contracts/solana/programs/allways_swap_manager/src/instructions/bind_hotkey.rs
```diff
@@ -0,0 +1,45 @@
...
+    let now = Clock::get()?.unix_timestamp;
+    let miner = ctx.accounts.miner.key();
+    let bump = ctx.bumps.binding;
+
+    let binding = &mut ctx.accounts.binding;
+    if binding.miner == Pubkey::default() {
+        binding.miner = miner;
+        binding.bump = bump;
+    }
+    binding.hotkey = hotkey;
```
LandynDev, line 63: **A5 seal — re-bind immutability is validator-policy, not endpoint-enforced.**

Re-bind overwrites `hotkey`/`hotkey_sig` in place, so the first-seen pin that makes the A1 strike-out cost something lives entirely off-chain. The strike-dodge vector: a struck pubkey rotates to a fresh pubkey (new `MinerState`, zeroed counters) and re-binds the *same* hotkey. Under the offload-to-contract principle, this is exactly the kind of checkable invariant the program could own rather than relying on every validator implementing the pin identically. See the reverse-PDA note on the `Binding` struct — a hotkey-keyed set-once marker would make the pin runtime-enforced. Not a correctness bug as written; flagging that the anti-sybil guarantee is softer than “baked into the endpoint.”

smart-contracts/solana/programs/allways_swap_manager/src/state.rs
```diff
@@ -246,12 +260,58 @@ pub struct MinerQuote {
...
+    pub bump: u8,
+}
+
+/// Per-miner identity binding (`seeds = [BIND_SEED, miner]`): links a miner's Solana pubkey to its
+/// Bittensor hotkey. `hotkey_sig` is an sr25519 signature by the hotkey over the miner's Solana pubkey;
+/// the contract only STORES it (sr25519 verify is too costly on-chain) — the validator verifies it
+/// off-chain and enforces 1:1 + first-seen pinning. Overwritable in place (the validator pins policy).
+#[account]
+#[derive(InitSpace)]
+pub struct Binding {
```
LandynDev, line 296: **A5 1:1 — the reverse direction isn’t enforceable from this PDA alone.**

Keyed `[BIND_SEED, miner_pubkey]`, so pubkey→≤1 hotkey holds structurally. But hotkey→≤1 pubkey doesn’t: two distinct pubkeys can each bind the same hotkey, and only the off-chain first-seen pin rejects the second. In theory this is the single seal that closes strike-dodge fully, and it *can* be on-chain: a second, hotkey-keyed marker PDA (`seeds = [<HOTKEY_BIND_SEED>, hotkey]`, `init`-only / set-once). The first pubkey to claim a hotkey creates it; any later pubkey trying the same hotkey fails the `init` at the runtime level → first-seen pin enforced by the program, not by validator agreement. One tiny rent-funded PDA per identity — same bounded family as the per-direction stats, not per-event. Offered as the cleaner “contract owns what’s checkable” shape; the off-chain pin still works, it’s just softer and divergence-prone.

LandynDev (commented):
## Plan-vetting pass on the A1–A5 read-surface (contract side)

Walked the whole batch against the offload-to-contract / “contract owns what’s checkable” principle. **Net: the design is in good shape** — a few items flagged for one more round, all explanatory rather than blocking. Inline notes on the changed files; the activate/quote gate is below since those files are untouched by this PR.

**Confirmed clean / nice simplifications:**
- **A3** — the front-run + squat surface is closed several ways over: validator-relayed claim (`ensure_validator`), payout pinned in the immutable `Reservation` (read from the reservation, never from claim args), one-claim-per-reservation (`claimed_swap_key`), plus `close_stale_claim` as the permissionless reaper. The original permissionless-claim concern (front-run with a wrong `user_to_addr`) is fully mooted. Relay-only (vs also user-direct) reads as a clean UX call.
- **A4 / A6** — source `TxMarker` removed in favor of the `Reservation.created_at` freshness floor, and there’s correctly no dest marker: both legs are now symmetric validator freshness checks (source vs `created_at`, dest vs `initiated_at`). This dissolves the unbounded-permanent-state cost (the marker was ~1 SOL / 1k txs, never closed) while keeping replay defense. Folding A6 into A4 was the right call.

**Flagged for discussion (see inline + below):**
1. **A2 `total_sol_amount`** (inline, state.rs) — may be redundant and bakes in collateral=SOL; the per-direction from/to totals are already asset-pure and the validator normalizes off-chain anyway.
2. **A5 binding seals** (inline on `bind_hotkey.rs` + `Binding` struct) — re-bind immutability and the reverse hotkey→pubkey 1:1 are currently validator-policy rather than endpoint-enforced; both can be sealed on-chain with one hotkey-keyed set-once marker PDA. This is the guarantee that most wants to live on-chain, since it’s what makes the A1 strike-out actually cost something.
3. **A5 binding-gate (seal that touches untouched files, noted here):** neither `vote_activate.rs` nor `set_quote.rs` requires a `Binding` to exist, so an **unbound pubkey can activate, quote, win pool draws, and run swaps** — its identity only resolves later, off-chain, when the validator maps state→UID. If a binding is what ties a pubkey to a registered miner, a one-line `require!` (the `Binding` PDA exists / has a non-default hotkey) at each endpoint keeps the invariant in the contract instead of as an off-chain filter, and closes the “operate-while-unbound” edge where a pubkey accrues stats/counters no UID will ever claim. `vote_activate` is the higher-value one; `set_quote` is lower priority since a quote alone earns nothing. Defense-in-depth, not urgent.

**Forward note for Phase B (validator side, not this PR):** the freshness check compares an external tx’s block timestamp to a Solana `unix_timestamp` — both unix secs, so no unit mismatch, but BTC block timestamps are loose (the +2h future rule + ~1h median-time-past lag, non-monotonic block-to-block) and Solana’s clock is itself a stake-voted clock that can drift. A strict `>` risks false-rejecting an honest *recent* deposit; a grace margin (or comparing against BTC `mediantime`) keeps replay defense intact — replayed txs are old by far more than any grace — while avoiding the false-reject. Worth baking into the freshness helper.

None of these block merge on their own — posting so the next round has the full picture.

PR #517, opened 2026-07-08
feat: restore deferred-confirmation swap intake — validator reaper + CLI post-tx relay

anderdc:
## What & why

Restores the ink!-era **deferred confirmation** for source deposits, dropped in the Solana port (#496). Previously `confirm_deposit` rejected a source tx unless it was *already* fully confirmed — fine for SOL (~13s) but it forced a BTC taker to wait ~20 min for 2 confs before `post-tx` even succeeded, timing out honest BTC takers who were just waiting on blocks.

**Model — the contract is the queue** (no validator-side queue; the old ink! `pending_confirms` SQLite table stays gone): a Swap PDA in `PendingAttestation` stores `from_tx_hash`, and the crank re-checks it each round (`pending`→extend, `ok`+fresh→attest). Voting is deferred, not the claim.

### Validator — deferred intake + stale-claim reaper (`4cac166`)
- **`reserve_engine.confirm_deposit`**: accept a content-valid deposit even before it fully confirms. Fast-fail with **no claim** only when `verify_transaction` returns `None` (absent/content-mismatch), so the short TTL frees the miner. Freshness is enforced only when `block_time` is present (mined); a 0-conf mempool tx defers freshness to the crank's `ok` gate (BTC Esplora returns full vin/vout for mempool txs).
- **`solana_swap_loop`**: new `CANCEL` decision + `_claim_is_stale` mirroring the contract's `close_stale_claim` guard — reaps an orphaned `PendingAttestation` claim (reservation expired past its ceiling, or slot re-resolved), freeing the miner + reclaiming rent. Closes the deadline gap: `decide()` previously had no timeout branch for `PendingAttestation`. Benign first-wins race handling.
- **`solana/{client,layouts}.py`**: wire the **pre-existing** `close_stale_claim` instruction (discriminator + builder). Its event was already mapped in #514; only the trigger was missing. No contract/IDL change.

### CLI — re-port `alw swap post-tx` (`4bdd575`)
- `alw swap post-tx <tx-hash>` relays the source-tx hash to serving validators via `SwapConfirmSynapse`; each validator verifies the on-chain deposit against the pinned Reservation and submits the claim. Permissionless — signed by an ephemeral throwaway hotkey as transport only (the on-chain deposit is the auth), so a pure-Solana taker needs no registered neuron. Resolves the winning miner from the pending-swap stash (fast path) or by scanning bindings; `--miner` relays on anyone's behalf. Verified live on devnet/testnet.

## Not a bug (documented)
`Reservation.claimed_swap_key` reading all-zeros post-attestation is by design — `vote_initiate` zeroes it + `reserved_until` at quorum. The double-claim guard is set at `submit_swap_claim` and cleared at attestation.

## Config (companion, separate PR)
- **reservation_ttl 2500→480** (8-min anti-grief gate; the conf runway is carried by extensions): deploy default in **entrius/alw-utils#84**; the live cluster needs a one-time `alw admin set-reservation-ttl 480` (Config PDA persists across redeploys).
- **max_total_extension stays 7200** — 8400 would exceed the hard on-chain lid (`MAX_TOTAL_EXTENSION_SECS_MAX`) and require a redeploy; at ttl=480 the ~128-min source ceiling already drives block-variance BTC 2-conf failures to ~1-in-26k.

## Testing
- **640 passed.** +18 new: `confirm_deposit` accepts unconfirmed-mempool / mined-low-conf / deeply-confirmed (fast-chain regression), rejects absent-or-mismatch / stale-mined / expired / already-claimed / unreachable — all without a claim; crank reaper decisions (incl. source-pending-but-expired) + live-matching regression; `close_stale_claim` cast + benign races; builder discriminator verified vs Anchor `sha256("global:…")` + account metas.
- CLI `post-tx` path verified live on devnet/testnet (ok → on-chain Swap created → FULFILLED).
- **Not run here:** validator-side devnet + BTC testnet4/signet integration — an 8-min BTC-source swap surviving via extension to 2 confs, and a dropped-tx reap.

@landyndev — swap CLI + validator swap lane.

### DIFF
--- a/allways/cli/swap_commands/helpers.py
+++ b/allways/cli/swap_commands/helpers.py
@@ -10,6 +10,7 @@ from typing import Callable, List, Optional, Tuple
 import bittensor as bt
 import click
 import requests
+from bittensor.utils import ss58_encode
 from rich.console import Console
 from rich.text import Text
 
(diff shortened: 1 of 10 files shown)


anderdc:
Companion deploy-default change (fresh clusters init at reservation_ttl=480): entrius/alw-utils#84

PR #544, opened 2026-07-12
perf(validator): hoist the BTC chain tip out of the per-leg confirmation loop

anderdc:
Port of the shipped **SOL** tip hoist (#542 / #543) to the **Bitcoin** provider — the same "fetch the tip once per pass, not once per leg" change, one function.

## What
`api_calc_confirmations` fetched `/blocks/tip/height` itself on every call, so tip fetches scaled with the number of BTC legs per forward pass even though the answer is identical for all of them. It now reads the base `cached_block_height()`.

```python
tip_height = self.cached_block_height()
if tip_height is None:
    return 0
return max(0, tip_height - block_number + 1)
```

## Why it's safe (inherited, not re-derived)
- `clear_provider_caches` **already** calls `clear_pass_tip()` on every provider each pass, so the validator now issues **one `/blocks/tip/height` per pass** instead of one per mined leg.
- The **15s TTL from #543 lives in the base** `cached_block_height`, so the non-clearing consumers that bit SOL (miner fulfillment, validator axon-reserve) are already protected for BTC — the tip **can never freeze**.
- A stale-low tip only biases confirmations **low** (conservative, never a false confirm); a failed fetch returns `None` (uncached) → `0` confirmations, same as before.

## Scope
**Tip hoist only.** The per-block reorg check (`/block/{hash}/status`) and the immutable tx-body cache (`/tx/{txid}`) are separate follow-ups — untouched here.

## Effect
Average day (~15 BTC legs): **~45 → ~31 calls/pass (~30%)**; also kills the per-leg tip in the max-burst case.

## Tests
`tests/test_btc_tip_hoist.py` mirrors the SOL tests: confirmations math, **one fetch per pass** across N legs, `clear_pass_tip` re-fetch, and failed-fetch → 0 (uncached). Full run: 37 passed (new + existing SOL/BTC provider suites).

### DIFF
--- a/allways/chain_providers/bitcoin.py
+++ b/allways/chain_providers/bitcoin.py
@@ -179,15 +179,16 @@ class BitcoinProvider(ChainProvider):
     # The sole data source for tx verification, confirmations, and balances.
 
     def api_calc_confirmations(self, block_number: int) -> int:
-        """Fetch the chain tip from Esplora and calculate confirmations for a block."""
-        try:
-            tip_resp = self.btc_api_get('/blocks/tip/height', timeout=10)
-            if tip_resp.ok:
-                tip_height = int(tip_resp.text.strip())
-                return tip_height - block_number + 1
-        except Exception:
-            pass
-        return 0
+        """Confirmations for a mined block from the pass-cached chain tip, so N BTC legs in one
+        forward pass share a single ``/blocks/tip/height`` instead of one each (mirrors the shipped
+        SOL hoist, #542/#543). The validator clears the tip each pass (one fetch/pass); the base's
+        15s TTL caps staleness for callers that never clear (miner fulfillment, axon-reserve), so the
+        tip can never freeze. A stale-low tip biases confirmations low — conservative, never a false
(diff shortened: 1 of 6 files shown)


anderdc:
Added the F2 CLI crank-race fix from the 2026-07-12 mainnet pressure test (aea230c).

Bug: _self_crank_resolve treated a lost resolve_pool race as benign only when the error carried the Anchor error name (pre-flight simulation). A tx that submits and lands failed surfaces through the confirm path as {'InstructionError': [0, {'Custom': 6044}]} — numeric code only, no name — so it slipped the match, re-raised, and aborted swap now, abandoning the taker's already-paid, since-drawn seat (0.02 SOL fee forfeited + miner locked busy_until ~10 min). Most frequent failure in the pressure run.

Fix: match both the name and the numeric code (6042 PoolNotClosed / 6044 NoRequests / 6045 SeedSlotNotYetProduced / 6046 AlreadyFilled). No contract change. 3 new tests. Bundled here per Alex; PR stays open while we investigate F1. Context in MAINNET_SWAP_OBSERVATIONS.md.

PR #549, opened 2026-07-13
feat(contract): floor posted rates to 5 sig figs to end the sub-perceptible crown war

anderdc:
## Problem

The crown is ranked **off-chain** on the raw stored rate — strict best-rate-wins, bucketed by exact rate value, ties split evenly (`scoring.py` `crown_holders_at_instant`). `set_quote` stored a **full-precision `u128`** with no tick (`state.rs` comment: "deliberately no on-chain rate check").

So two miners could undercut in a **sub-perceptible digit** (e.g. the 9th sig fig) to land in a distinct, strictly-better bucket and capture the **entire** crown for a direction — for free, with **zero benefit to takers**. A rate war with no floor. The only friction was the anti-flashing churn fee, which taxes *rapid* re-quotes but doesn't stop a slow, invisible undercut.

`RATE_SIG_FIGS = 5` and `normalize_rate()` already existed and *claimed* (in comments) to be enforced on post and parse — but neither the CLI post path nor the validator ingest actually applied it. It was display-only.

## Fix

Floor every posted rate to `RATE_SIG_FIGS` (=5) significant figures **at the source**, so a crown-winning improvement is always one a taker can actually see; sub-perceptible undercuts collapse into the incumbent's bucket and **tie & split** instead of stealing the crown.

- **Contract** — `quantize_rate_sig_figs` (pure integer `ilog10` + floor, no floats — non-deterministic in BPF; floor not round so no tick-gaming and no overflow), applied in `set_quote` **before the rate is stored *or* emitted**, so the pinned swap rate, off-chain crown ranking, and indexer/UI all read one canonical value.
- **CLI** — mirror the floor (`quantize_rate_fixed`) so previews and posts match exactly what the chain stores.
- **Validator** — floor on `QuoteSet` ingest too. Redundant post-redeploy, but **closes the migration window** where a pre-redeploy full-precision quote could still snipe the crown before it re-posts.
- Fix the stale `RATE_SIG_FIGS` comment that claimed enforcement already existed.

## Decisions

- **Hardcoded constant, not a `Config` field** — `Config` is a live singleton account ending in `bump: u8`, so adding a field is a borsh layout migration on the existing account. Keeping it a const keeps the redeploy free of an account migration. Can be promoted to `Config` later if we ever want to retune the tick without a program upgrade.
- **No IDL / client change** — `set_quote`'s signature (`rate: u128`) and the `QuoteSet` event shape are unchanged; only the stored value is floored. The alw-utils indexer needs no change.
- **5 sig figs preserved** from the existing constant — not repicking precision here.

## Test / verify

- Rust: `cargo test -p allways_swap_manager --lib quantize` → ok.
- Python: `pytest tests/test_rate.py` → 62 passed (includes new `TestQuantizeRate`, kept in lockstep with the Rust cases).
- Import smoke on validator + both CLI modules.

## Rollout

Requires a **contract redeploy** (testnet netuid19 → mainnet netuid7) for the on-chain floor to take effect. The validator-side ingest floor and CLI mirror take effect on normal deploy and already neutralize the war for the crown independent of the contract, so the redeploy is about canonical storage + execution/UI consistency, not a hard dependency for stopping the bleed.

### DIFF
--- a/allways/cli/swap_commands/numeraire.py
+++ b/allways/cli/swap_commands/numeraire.py
@@ -30,6 +30,7 @@ from allways.cli.swap_commands.helpers import (
 from allways.cli.swap_commands.pair import write_rate_posted_flag
 from allways.constants import LAUNCH_SPOKES, NUMERAIRE_CHAIN, RATE_PRECISION
 from allways.solana.client import SolanaClientError
+from allways.utils.rate import quantize_rate_display, quantize_rate_fixed
 
 
 @dataclass
@@ -149,7 +150,8 @@ def quotes_command(spread_bps, dry_run, yes, **spoke_opts):
                 )
             else:
                 note = f'[dim]free (was {was:g})[/dim]'
-        console.print(f'  {sp.from_chain.upper()} → {sp.to_chain.upper()}: [green]{sp.rate:g}[/green]   {note}')
+        disp = quantize_rate_display(sp.rate)  # mirror the on-chain floor so preview == stored
+        console.print(f'  {sp.from_chain.upper()} → {sp.to_chain.upper()}: [green]{disp:g}[/green]   {note}')
 
     if total_fee:
         console.print(
@@ -168,7 +170,14 @@ def quotes_command(spread_bps, dry_run, yes, **spoke_opts):
     for sp in specs:
         try:
             with loading(f'Publishing {sp.from_chain.upper()} → {sp.to_chain.upper()}...'):
-                client.set_quote(sp.from_chain, sp.to_chain, sp.from_addr, sp.to_addr, int(sp.rate * RATE_PRECISION), 0)
+                client.set_quote(
+                    sp.from_chain,
+                    sp.to_chain,
+                    sp.from_addr,
+                    sp.to_addr,
+                    quantize_rate_fixed(int(sp.rate * RATE_PRECISION)),
+                    0,
+                )
             posted += 1
         except SolanaClientError as e:
             console.print(f'[red]Failed {sp.from_chain.upper()} → {sp.to_chain.upper()}: {e}[/red]')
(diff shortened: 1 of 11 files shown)


anderdc:
Folded in a second, independent contract change (commit \`6d8ee77\`) so it rides the same redeploy — flagging it since it's outside the rate-quantization scope of the original diff.

**\`fix(contract): close per-swap vote rounds on quorum to reclaim their rent\`**

\`confirm_swap\`/\`timeout_swap\` already close the Swap PDA in the terminal quorum branch but only \`reset_round\`'d the per-swap \`VoteRound\`, parking ~0.00482 SOL/swap on-chain forever. Now the round is closed too, refunding its rent to the validator.

- **Scope:** the two per-swap round types (\`VOTE/CONFIRM/swap_key\`, \`VOTE/TIMEOUT/swap_key\`) are the only *unbounded* rounds — initiate/activate/deactivate are per-miner, set_weights is a global singleton, all reused/reset in place. So this fully stops the single-validator leak; nothing else stacks.
- **Straggler safety:** \`swap\` is a required \`Account<Swap>\` resolved *before* \`vote_round\` in the accounts struct, so a late confirm/timeout reverts on the already-closed swap before \`init_if_needed\` could re-create the round. Worst case is rent, never funds — and this path already reverted today (the swap close is pre-existing); this commit only adds the rent refund.
- **Tests:** the existing multi-validator LiteSVM lifecycle tests now assert the confirm/timeout round is gone after quorum. Full suite green (rebuilt the BPF \`.so\` first — stale-\`.so\` trap).

**Deliberately *not* included:** the backlog reaper instruction + off-chain sweep (Part 2/3 of the original plan). Its real job is multi-validator losing-branch orphans, which are only produced/testable once multi-validator ships — and that work brings its own contract upgrade, the reaper's natural home. On single-validator the backlog (~33 test-artifact orphans, ~0.16 SOL) is frozen by this fix and won't grow. Will track it against the multi-validator milestone.

PR #562, opened 2026-07-15
Add validator stake-weight vote to Solana contract

LandynDev:
## What

Python side of the stake-weight mechanism, plus one small contract change: every ~12h (block-aligned, 3600 subtensor blocks) each validator derives the draw-weight vector — floor(alpha_stake / 50k) per whitelisted validator, index-aligned to Config.validators — and votes it via vote_set_weights. Quorum applies it; the reservation-lottery draw becomes stake-true.

## How

- layouts.py / client.py: vote_set_weights builder targeting the post-#560 interface — args (weights: Vec<u64>, round_key: [u8;32]), round_key = keccak(REQ_SET_WEIGHTS || validator keys || weights LE) via weights_round_key (mirror of consensus.rs::weights_hash), vote-round PDA keyed by that hash (per-snapshot rounds)
- validator/weights_vote.py: block-boundary gate, in-epoch retry throttle (matches vote-round TTL), skip-if-unchanged, per-snapshot open-round dedup, benign-error swallowing (AlreadyVoted / WeightsUpdateTooSoon / NotValidator); never raises into the forward loop
- Not-whitelisted validators warn once and keep running; watch mode logs WOULD
- Contract (bind_hotkey.rs): the H3 squat gate now also admits whitelisted validators — miner_state becomes an optional account; the gate is collateralized-MinerState OR signer in Config.validators. Without this, validators cannot bind and every derived weight is 0. Squat pricing holds: both paths cost something scarce and attributable (locked collateral / hand-approved slot)
- bind-hotkey is now a top-level alw command (miners and validators; alw miner bind-hotkey kept as alias); validator startup warns if unbound; the client passes anchor's optional-account None when the signer has no MinerState
- constants.py: WEIGHTS_STAKE_BUCKET_ALPHA, WEIGHTS_VOTE_INTERVAL_BLOCKS, vote-round-TTL mirror

## Testing

- Python: 743 passed (26 new: vector derivation edges, every step gate, error swallowing, instruction bytes incl. independent keccak recompute of round_key)
- Contract: all 12 suites green; new test_bind_whitelisted_validator_without_miner_state + test_bind_none_miner_state_still_gated
- Solana e2e (fresh deploy of this branch's program): validator binds via the widened gate, bootstrap vote lands, Config shows the applied vector + last_weights_update stamped; sol-btc lifecycle unaffected

## Deploy note

Contains a program change — requires a contract redeploy/upgrade alongside the validator rollout.


LandynDev:
Revision pushed: the builder now targets the post-#560 interface — args are (weights: Vec<u64>, round_key: [u8;32]), round_key = keccak(REQ_SET_WEIGHTS || validator keys in config order || weights u64 LE) via a new weights_round_key helper mirroring consensus.rs::weights_hash, and the vote_round PDA is keyed by that hash (per-snapshot rounds). weights_vote.py derives round_key from the exact snapshot it submits; VoteHashMismatch dropped from the benign-error list (unreachable under per-snapshot rounds). Builder test recomputes the hash independently and asserts the new arg layout + keyed PDA. Verified on-chain against a freshly deployed post-#560 program in the solana e2e: the bootstrap vote lands and Config shows the applied vector + last_weights_update stamped. 743 tests green.

LandynDev:
Second revision pushed: #560's bind gate blocked validators from binding at all (no MinerState), which would have made every derived weight 0. bind_hotkey's miner_state is now optional and the gate admits whitelisted validators (collateral OR whitelist slot — squat pricing preserved). Rust tests added for the validator path and the None-without-whitelist rejection. NOTE: the PR now contains a program change — needs a contract redeploy when it ships. E2E on a fresh deploy of this branch: validator binds, bootstrap vote lands, weights applied on-chain.

LandynDev:
Simplification pass: dropped the open-round pre-check (_voted_in_open_round + get_vote_round read + the TTL-mirror constant) — the contract's AlreadyVoted rejection plus the 30-min retry throttle already cover it, and at this cadence the pre-check saved nothing (a duplicate fails preflight, fee-free). Also dropped the once-per-process whitelist-warning flag (the epoch memo already rate-limits it to once per ~12h). Net -44 lines. 741 tests green; e2e re-run PASS (bound, weights=[2] posted, lifecycle COMPLETED).

anderdc (approved):

PR #606, opened 2026-07-28
Verify TAO transfers by settlement rather than by the decoded call

anderdc:
Small correctness fix in the TAO chain provider.

## What

`fetch_matching_tx` and `find_recent_outgoing` matched on the decoded `Balances.transfer` call. On Substrate a signed extrinsic that is included in a block but dispatches with an error (insufficient balance, for one) still decodes to the intended dest and amount, so it satisfied the match without having moved any funds. Inclusion is not settlement.

Both paths now require a `Balances.Transfer` event emitted by that extrinsic before treating it as a deposit, and read the amount and sender from the event rather than from the call.

Supporting changes:

- Extrinsic index is preserved through the raw-block path, since events key on position within the full block.
- Unreadable events raise `ProviderUnreachableError` rather than returning "no match". Callers map that to skip/retry; a plain `None` is slash-eligible upstream, so a node hiccup must not land there.
- Block hashes and per-block events are cached alongside the existing block cache and cleared on the same `clear_cache` cycle, so this adds at most one extra storage read per block touched.
- `parse_raw_extrinsic` read the signer at `body[1:33]`, straddling the MultiAddress variant byte and producing a shifted (valid-looking but wrong) ss58. Corrected to `body[2:34]`; non-Id signer variants now fail closed. Note this only affected the raw fallback path — the primary path takes the sender from the decoded extrinsic.

## Scope

TAO only. The Solana provider already gates on `meta.err` and reads pre/post balance deltas; the Bitcoin provider reads vouts of a confirmed tx. Both verify effect. Substrate is the one backend of the three where inclusion and effect can diverge, so the shared `fetch_matching_tx` shape carried an assumption that does not hold there.

## Tests

`tests/test_tao_settlement.py` (new) covers the by-hash verifier: included-but-unsettled returns None, amount and sender come from the event, settlement to another recipient or below the asking amount does not match, overpayment still settles, and unreadable events raise rather than read as absent.

`tests/test_block_time.py` extends the scanner cases with the unsettled and short-settled variants, `settled_credit` unit cases, and the extrinsic-decode fix. The event-shape normaliser is parametrised over the shapes scalecodec emits across runtime versions.

825 passed, ruff clean.

## Before merge

The event-shape normaliser was written against the `get_events()` contract and covered by parametrised tests, but I could not reach a websocket endpoint to confirm the exact dict shape this runtime returns. Worth one smoke test against a live node — verify a known-good TAO leg still resolves — before this goes out.

### DIFF
--- a/allways/__init__.py
+++ b/allways/__init__.py
@@ -1,3 +1,3 @@
-__version__ = '2.0.2'
+__version__ = '2.1.0'
 version_split = __version__.split('.')
 __spec_version__ = (1000 * int(version_split[0])) + (10 * int(version_split[1])) + (1 * int(version_split[2]))
--- a/allways/chain_providers/subtensor.py
+++ b/allways/chain_providers/subtensor.py
@@ -307,15 +307,22 @@ class SubtensorProvider(ChainProvider):
         if not block_hash:
             raise ProviderUnreachableError(f'TAO block hash unavailable for {block_num}')
 
+        # Reached only once an extrinsic was located in this block, so the block provably holds one
+        # and must emit its ApplyExtrinsic records. An empty list is a broken response, not an
+        # empty block, and reporting it as no-credit would false-slash on an RPC hiccup alone.
+        events = self.get_block_events(block_hash)
+        if not events:
+            raise ProviderUnreachableError(f'no events returned for block {block_num}, which holds extrinsics')
+
         credited = 0
         sender = ''
         indexed = 0
-        events = self.get_block_events(block_hash)
         for record in events:
-            if self._event_extrinsic_idx(record) is None:
+            idx = self._event_extrinsic_idx(record)
+            if idx is None:
                 continue
             indexed += 1
-            if self._event_extrinsic_idx(record) != extrinsic_idx:
+            if idx != extrinsic_idx:
                 continue
             transfer = self._transfer_from_event(record)
             if transfer is None:
@@ -326,10 +333,8 @@ class SubtensorProvider(ChainProvider):
             credited += ev_amount
             sender = sender or ev_sender
 
-        # Any block holding extrinsics emits ApplyExtrinsic records, so recognising none of them
-        # means the phase shape moved, not that the block is empty. Same reasoning as above: that
-        # is 'unknown', and reporting it as no-credit would false-slash on shape drift alone.
-        if events and not indexed:
+        # Recognising none of them means the phase shape moved, not that nothing was applied.
+        if not indexed:
             raise ProviderUnreachableError(
                 f'no ApplyExtrinsic phase recognised in {len(events)} events at block {block_num}'
             )
--- a/allways/chain_providers/subtensor.py
+++ b/allways/chain_providers/subtensor.py
@@ -255,6 +255,12 @@ class SubtensorProvider(ChainProvider):
                 return None
             attributes = balances['Transfer']
 
+        # Past this point the record IS a Balances.Transfer, so a payload we can't read is
+        # 'unknown', never 'absent'. Returning None here would read a real deposit as moving
+        # no funds — on a dest leg that is slash-eligible, so shape drift would false-slash.
+        def unreadable(detail: str) -> ProviderUnreachableError:
+            return ProviderUnreachableError(f'unreadable Balances.Transfer payload ({detail}): {attributes!r:.200}')
+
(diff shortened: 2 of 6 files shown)


LandynDev (dismissed):

anderdc:
Good catch — applied in 30a7048, and it turned out to have a sibling.

Once a record is identified as a `Balances.Transfer`, an unreadable attributes payload now raises `ProviderUnreachableError` rather than returning `None`, so shape drift reads as unknown and callers skip and retry. The strictness starts only after the module/event match, so a non-transfer event with an odd payload is still just absent.

`settled_credit` had the same conflation one level up, via `phase` rather than `attributes`: if the phase shape moves, `_event_extrinsic_idx` returns `None` for every record, nothing matches the index filter, and the result is an empty credit — identical false-negative, identical false-slash path. Any block holding extrinsics emits `ApplyExtrinsic` records, so recognising none of them now raises as well. A genuinely empty event list is still an ordinary no-credit.

Tests added for both: the payload case is parametrised over renamed keys, short positional params, non-numeric amounts and unresolved accounts, plus the phase-drift case and a check that an empty block is not mistaken for drift. The unsettled-deposit test now uses realistic failed-dispatch events (`Balances.Withdraw` + `TransactionPayment.TransactionFeePaid` + `System.ExtrinsicFailed`) instead of an empty list, so it exercises the "events present, no Transfer" path rather than the trivial one.

835 passed, ruff clean.

LandynDev (dismissed):

anderdc:
Cleared the "before merge" caveat — validated the event shape against mainnet rather than deferring it.

Websockets are blocked from where I am, but the decode itself does not need them: fetched `state_getMetadata` + the `System.Events` storage over HTTPS JSON-RPC and decoded `scale_info::20` with the same scalecodec portable-registry path `get_events()` uses.

Real record shape, block 8723231:

```
record keys : [phase, extrinsic_idx, event, event_index, module_id, event_id, attributes, topics]
Balances.Transfer attributes: {"from": "5EYC...", "to": "5G68...", "amount": 12191619303}
```

`extrinsic_idx` is present and the attributes are the flat from/to/amount dict, which is the primary path in the normaliser. Ran both directions end-to-end through the real provider methods:

- **Negative** — block 8723231, the 4 TAO source leg of swap `5078723637387828482`: `fetch_matching_tx` returns `None`, logging `is in block 8723231 but moved no funds ... (dispatch failed) — rejecting`.
- **Positive** — block 8721438, a genuine 1.750034889 TAO deposit: accepted, `amount=1750034889`, sender read from the event as `5H3Symui...`.

Also a useful negative result on the new strictness: across 779 real event records in those two blocks (298 of them `Balances.Transfer`), nothing raised. So the fail-closed path does not trip on live data — `_event_extrinsic_idx` correctly returns `None` for the `Initialization`/`Finalization` records, which are emission and staking transfers emitted outside any extrinsic, and those are skipped rather than misread.

One honest caveat: this decodes the storage directly, whereas `get_events()` goes through `query("System", "Events").value`. Both land on the same scalecodec object, so the `.value` structure is the same, but it is not literally the same call. Cheap to confirm on a validator host if you want belt and braces.

anderdc:
Both applied in 615864b.

**1. Empty event list.** You are right, and it was the last door open. `settled_credit` is only reached once `fetch_matching_tx` or the scanner has already located an extrinsic in that block, so the block provably holds one and must emit its `ApplyExtrinsic` records — an empty list cannot mean "nothing settled". It now raises. Consistent with what I saw decoding live blocks: 404 events at 8723231 and 375 at 8721438, never empty, matching your 25-block sample.

That makes the rule uniform across all three: unreadable payload, unrecognised phase, and empty list all raise; only a well-formed set of events that genuinely contains no matching `Balances.Transfer` returns `None`.

The scanner fixture also modelled a failed dispatch as `_events: []`, which is not what a failure looks like. It now carries the `Balances.Withdraw` + `System.ExtrinsicFailed` records a real one leaves behind, so that test exercises the events-present path rather than the empty-list one it was about to start asserting.

**2. Duplicate call.** Lifted into a local — one call per record instead of two.

Re-ran the mainnet checks after the change: block 8723231 still rejects the unsettled 4 TAO leg, block 8721438 still accepts the genuine 1.750034889 TAO deposit with the sender read from the event. 835 passed, ruff clean.

LandynDev (approved):

PR #610, opened 2026-07-30
Return the quote rejection reason from seam /rate

LandynDev:
Seam /rate collapsed every rejection into "no executable quote for that pair/amount". A taker swapping 0.1 TAO->SOL on testnet (SOL leg 0.083, under the 0.1 SOL min_swap) saw the same message as a direction nobody quotes — indistinguishable from an outage.

Adds unviable_reason() beside viable_intakes() in swap_intake.py — same candidates, same gates, but it reports which one failed (direction unquoted / rate not executable / below min swap / above max swap / miner collateral too low). The seam's 404 body now carries that reason via quote_rejection_reason(); the happy path is untouched (best_quote's tail split into _best_quote_result, no behavior change).

Tests: below-min (the exact 0.1 TAO case), no-direction, and collateral reasons; full test_swap_intake + test_seam_http suites pass (28) in the entrius/allways:test image.

### DIFF
--- a/allways/cli/swap_commands/swap_intake.py
+++ b/allways/cli/swap_commands/swap_intake.py
@@ -122,6 +122,38 @@ def viable_intakes(
     return out
 
 
+def unviable_reason(
+    candidates: List[MinerCandidate],
+    from_chain: str,
+    to_chain: str,
+    from_amount: int,
+    min_swap: int,
+    max_swap: int,
+) -> str:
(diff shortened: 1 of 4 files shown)


LandynDev:
Added source-asset translation: bound reasons now read in what the taker is sending — `below min swap (≈0.12 TAO (0.1000 SOL leg))` for tao→sol, exact `0.1000 SOL` when SOL is the source. Display-only float conversion at each candidate's canonical rate, never fed into a gate. 29/29 in the :test image.

PR #611, opened 2026-07-30
Phrase swap bounds in the taker's source asset

LandynDev:
Follow-up to #610: bound rejections read in what the taker is sending — `below min swap (≈0.12 TAO (0.1000 SOL leg))` for tao→sol instead of SOL-only; exact `0.1000 SOL` (no ≈) when SOL is the source. Same for above-max.

Conversion is display-only float at each candidate's canonical rate, never fed into a gate; swap_viable itself is untouched (its strings are the contract mirror, prefix-coupling pinned by test).

Tests: 29/29 (test_swap_intake + test_seam_http) in the entrius/allways:test image; ruff clean.

### DIFF
--- a/allways/cli/swap_commands/swap_intake.py
+++ b/allways/cli/swap_commands/swap_intake.py
@@ -82,17 +82,34 @@ def compute_intake_amounts(from_chain: str, to_chain: str, from_amount: int, rat
     return IntakeAmounts(collateral_amount=collateral_amount, from_amount=from_amount, to_amount=to_amount)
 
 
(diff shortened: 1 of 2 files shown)


LandynDev:
Refactored away the string-prefix coupling: gates now live in `swap_gate()` returning a verdict code ('' | below_min | above_max | low_collateral); `swap_viable` keeps its exact (bool, SOL-message) API as a thin formatter for the miner/CLI callers, and `unviable_reason` formats taker-facing text from the code directly. `viable_intakes` gates on the code (no string built on the hot path). 29/29 in the :test image, ruff clean.

PR #616, opened 2026-08-04
Collateral-floor sweep: drive vote_deactivate after a min_collateral raise

anderdc:
## What

Stacked on #614. The precondition for raising `min_collateral` (planned 0.5–1 SOL): without this, an admin floor raise strands every under-floor miner as a permanent zombie — `set_min_collateral` doesn't touch `MinerState`, `open_or_request` refuses to reserve them, so the fee/slash `apply_penalty` path that auto-deactivates under-floor miners can never fire. They stay `active` indefinitely, unreservable by takers but still earning crown. The contract's `vote_deactivate` quorum instruction handles exactly this case; it just had no driver — no discriminator in `layouts.py`, no client method, no validator loop.

## Cost model (the "be smart about RPC" part)

- **Steady state: zero added RPC.** The sweep reads `min_collateral` off the existing `SolanaConfigCache` (one Config read per 300s already paid for by swap bounds) and compares it to the last-seen value — one int comparison per forward step.
- **The `getProgramAccounts` scan runs only when the cached floor rises**, or once on the first step after boot (covers a raise that happened while this validator was offline). Floor decreases don't rescan.
- **Stragglers get per-miner account reads on a 300s retry cadence**, only while the pending set is non-empty: busy miners (the contract rejects kicking a miner with an in-flight swap / unexpired `busy_until` — they retry once idle; they can't take *new* swaps, so everyone converges within one fulfillment timeout), and cast votes awaiting quorum (a miner leaves pending only when the chain shows `active = false` or collateral ≥ floor, so a restart or reset round can't strand a half-kicked miner).
- `has_voted` gate prevents duplicate votes; a miner that tops up instead of getting kicked is released.

## Pieces

- `layouts.py`: `vote_deactivate` discriminator (sha256(`global:vote_deactivate`)[:8], derivation verified against the known `deactivate` discriminator).
- `client.py`: `vote_deactivate(miner)` mirroring `vote_activate` with `REQ_DEACTIVATE`.
- `bounds_cache.py`: `min_collateral()` accessor on the existing TTL cache.
- `floor_sweep.py`: the sweep + a never-raises `maybe_sweep_floor` forward hook (same containment pattern as `maybe_vote_weights`). Respects watch mode (`WOULD vote_deactivate …`).
- 8 new tests: steady-state costs nothing, raise kicks under-floor active idle miners only, busy-retry, no double-vote while awaiting quorum, top-up release, watch mode.

## Deploy order

Validators must run this **before** the `alw admin set-min-collateral` raise fires — sequence: merge → deploy to ≥67% of validator stake → raise the floor. The under-floor zombie window is then bounded by vote latency (minutes), not forever.

Full suite: 856 passed.

### DIFF
--- a/allways/constants.py
+++ b/allways/constants.py
@@ -114,6 +114,9 @@ RECONCILE_QUIET_SECS = 600
 SECONDS_PER_BLOCK = 12
 WEIGHTS_STAKE_BUCKET_ALPHA = 50_000  # alpha per draw-weight unit; floor rounding
 WEIGHTS_VOTE_INTERVAL_BLOCKS = 3_600  # ~12h — posting boundary cadence
+# Contract vote-round lifetime — mirrors constants.rs. A non-empty round older than this is
+# stale: record_vote clears and reopens it, and prior voters may legally vote again.
+VOTE_ROUND_TTL_SECS = 1_800
 # In-epoch retry throttle: one attempt per contract vote-round lifetime (VOTE_ROUND_TTL_SECS),
 # so an unlanded round has expired (and is reopenable with our snapshot) by the time we retry.
 WEIGHTS_VOTE_RETRY_SECS = 1_800
--- a/allways/solana/client.py
+++ b/allways/solana/client.py
@@ -595,6 +595,21 @@ class AllwaysSolanaClient:
         ]
         return self._send([self._ix('vote_activate', b'', metas)])
 
+    def vote_deactivate(self, miner) -> str:
+        """Vote to force-deactivate an idle active miner; on quorum active flips false.
+        The contract rejects busy miners (in-flight swap / unexpired busy_until)."""
+        validator = self.keypair.pubkey()
+        m = _as_pubkey(miner)
+        metas = [
+            AccountMeta(validator, True, True),
+            AccountMeta(pdas.config_pda(self.program_id), False, False),
+            AccountMeta(m, False, False),
+            AccountMeta(pdas.miner_state_pda(m, self.program_id), False, True),
+            AccountMeta(pdas.vote_round_pda(pdas.REQ_DEACTIVATE, m, self.program_id), False, True),
+            AccountMeta(SYSTEM_PROGRAM, False, False),
+        ]
+        return self._send([self._ix('vote_deactivate', b'', metas)])
+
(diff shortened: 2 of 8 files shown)


LandynDev:
Reviewed — one fix requested before merge.

`has_voted` (allways/solana/client.py:246) only checks voters membership, but the program reopens rounds older than VOTE_ROUND_TTL_SECS (1800s) in record_vote before the AlreadyVoted check — a prior voter can legally re-vote a stale round, and the client gate blocks exactly that.

Failure path: first vote pass misses quorum (e.g. partial rollout, 3/5 upgraded at 66% threshold) → round goes stale → every prior voter sees has_voted=True on each 300s recheck and skips forever. Recovery needs a vote from a validator outside the stale voter list; validator churn can leave none, permanently stranding the miner despite the sweep.

Fix options: treat a round with now - created_at > VOTE_ROUND_TTL_SECS as not-voted (created_at is already in the VoteRound layout), or drop the gate and treat the on-chain AlreadyVoted error as benign in _resolve.

Everything else checked out: discriminator recomputed and matches, account metas / PDA seeds / REQ_DEACTIVATE verified against the program, state machine probed (raise-while-pending, failed scans, deregistration, read-only, partial deploy), 152/152 tests pass.

anderdc:
Good catch — verified against `record_vote`: the stale-round clear does run before the `AlreadyVoted` check, so the client gate was blocking a legal re-vote and the partial-rollout scenario stranded the miner exactly as described.

Fixed in bbf105e, went with option (a) plus a belt-and-braces version of (b):
- `has_voted` now mirrors the contract rule: empty round or `now - created_at > VOTE_ROUND_TTL_SECS` (mirrored into `constants.py`) → not voted. This also fixes the same latent hole in the swap loop's `REQ_INITIATE`/`CONFIRM`/`TIMEOUT` skip guards, which share the helper.
- The sweep additionally treats a landed `AlreadyVoted` as a benign no-op for the read-then-send race, rather than a warning-level failure.

Kept the gate itself (vs dropping it entirely) so the steady await-quorum path stays a free account read instead of a failed tx + fee every 300s recheck. New tests cover the TTL boundary (live at exactly TTL, stale at TTL+1), the empty-round case, and the AlreadyVoted race.

LandynDev (approved):

PR #618, opened 2026-08-05
Harden ETH provider slash paths and send idempotency

LandynDev:
Surgical fixes for the ETH-spoke review findings (#1 smart-wallet slash exploit is parked pending design go-ahead). One file of production changes.

## Validator slash paths (false-slash prevention)
- **Absence quorum** (`eth_rpc null_needs_quorum`): a tx-lookup null is authoritative only when every endpoint reachably agrees; null + an unreachable endpoint raises `ProviderUnreachableError` (SKIP). One stale server in a public load balancer can no longer slash an honest miner.
- **Block-time unknown ≠ stale**: a failed block/timestamp fetch on a mined, settled tx raises (SKIP) instead of returning `block_time=None`, which the freshness gate fails closed on — previously a slash.

## Miner send correctness
- Broadcast success always returns the locally-computed tx hash — a `result: null` broadcast reply can no longer persist a blank hash that permanently bricks the swap's retry loop.
- `send_amount` probes its own prior broadcasts by hash (mempool-inclusive) before sending again; an unresolvable probe aborts instead of risking a double pay.

## Config / scanner
- Unknown `ETH_NETWORK` (typo) raises instead of silently defaulting to mainnet.
- Deposit scanner parks its cursor below an unreadable block instead of skipping it forever.
- ETH dust-floor comment corrected (rate sanity floor, not an economic guarantee).

Tests: 933 passed (19 new: quorum matrix, block-time unknown, send resilience, cursor parking, network guard).

### DIFF
--- a/allways/chain_providers/ethereum.py
+++ b/allways/chain_providers/ethereum.py
@@ -78,8 +78,10 @@ class EthereumProvider(ChainProvider):
         # (key-by-hash); a reorg changes it → miss → full refetch. Only fully-settled entries
         # (status 1 + timestamp) are cached — never pending/absent/reverted.
         self._settled_cache: OrderedDict[str, dict] = OrderedDict()
-        # Own broadcasts, tx_hash → (to, amount): send dedup scoped to this process (#461 class).
-        self.broadcasted_txids: Dict[str, Tuple[str, int]] = {}
+        # Own broadcasts, tx_hash → (to, amount, head at broadcast): send dedup scoped to this
+        # process and to SCAN_LOOKBACK_BLOCKS (#461 class — a later same-amount swap must never
+        # resolve to an earlier swap's consumed tx).
+        self.broadcasted_txids: Dict[str, Tuple[str, int, int]] = {}
         # Deposit-scanner head cursors, keyed per (from, to, amount) — see find_recent_outgoing.
         self.scan_cursors: Dict[Tuple[str, str, int], int] = {}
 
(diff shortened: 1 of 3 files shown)


anderdc (changes requested):
Two problems in the send dedup rewrite (ethereum.py:392-402), both regressions against the `find_recent_outgoing` path it replaces.

1. The reuse probe accepts any tx that exists on chain. The code it replaces required receipt status 1 ("A mined match can still have reverted — require receipt status 1"); this drops that. A prior broadcast that reverted still resolves via `eth_getTransactionByHash`, so the miner hands back a reverted hash, the validator rejects it as moving no funds, and the swap times out into a slash — and it does so on every retry, forever, because the entry is never cleared. This is exactly the reverting-destination case #619 is about. Require status 1 before reusing, and treat status 0 as not-reusable so a fresh attempt can proceed.

2. `broadcasted_txids` has no time bound, so reuse now matches across swaps. The old scan was bounded by `scan_cursors` plus a 25-block lookback, which meant a prior swap's tx fell out of range within about five minutes. The new dict is keyed only on (dest, amount) and lives for the whole process, so a second swap paying the same address the same wei — a repeat customer at an unchanged quote, or a taker doing it deliberately — gets swap #1's hash back. The miner marks fulfilled with a tx older than swap #2's reservation, freshness rejects it, and the miner is slashed while the user gets nothing. Record the head block alongside each broadcast and only reuse within `SCAN_LOOKBACK_BLOCKS`, which restores the old window while keeping the mempool-inclusive probe this PR adds.

Minor: the new park at ethereum.py:489 returns before the `_MAX_SCAN_CURSORS` trim at the end of `find_recent_outgoing`, so a key that parks on every call never gets evicted.

LandynDev (commented):
All three addressed in 30be65c:

1. Reuse now requires mempool-or-settled: a mined prior broadcast must show receipt status 1; status 0 clears the entry (moved no funds) so a fresh send proceeds; an unavailable receipt aborts the send rather than guessing.
2. Broadcast records now carry the head block at broadcast and age out after SCAN_LOOKBACK_BLOCKS — same window as the mined scan this replaced. A later same-amount swap can no longer resolve to an earlier swap's consumed tx. An unreadable head aborts the send (can't scope the window → don't send).
3. Cursor writes go through _set_cursor, so the parked-key path hits the same eviction as the end-of-scan path.

Tests cover each: settled reuse, reverted-clears-and-resends, outside-window prune, receipt-unavailable abort, head-unavailable abort.

anderdc (approved):

PR #652, opened 2026-08-12
Add Quant (QNT) on Ethereum as a launch spoke

LandynDev:
## ⛔ blocked: awaiting testnet deployment — DO NOT MERGE

QNT has **no deployment on Sepolia** (`eth_getCode` at `0x4a220E60…4675` returns `0x` on both
configured endpoints). `Erc20._token_contract` raises when a non-mainnet network has no
`TESTNET_TOKEN_CONTRACTS` entry, and validators construct **every** provider at boot
(`neurons/validator.py`, `required_chains=None`). Merging this adds `qnt` to `LAUNCH_SPOKES`, and a
**testnet validator would then fail to start at all** — boot-blocking, not a coverage gap.

**What unparks it:** a project-deployed minimal ERC-20 on Sepolia — symbol `QNT`, 18 decimals,
**immutable, with no pause and no blacklist** (an OpenZeppelin `ERC20Pausable` mock is immutable and
still wrong: pausing it would slash every honest miner on the pair, and it would fail the boot
falsifier) — pinned in `erc20.py::TESTNET_TOKEN_CONTRACTS` with a comment stating plainly that it is
a **project-deployed test token, not an issuer deployment**. No mock was deployed here: that work is
deferred out of this batch by decision, and no fail-soft boot path was added in its place.

## Base branch

Opened against **`refactor/erc20-refusal-interface` (#647), not `test`** — QNT implements neither
`isBlacklisted` nor `paused`, both **revert**, and pre-#647 `_refused_at` would propagate that out of
`delivery_refused` and defer every slash on the pair forever. Retarget to `test` after #647 merges.

## What changed

| File | Change |
|---|---|
| `allways/chains.py` | `CHAIN_QNT` + `SUPPORTED_CHAINS` entry |
| `allways/assets/qnt.py` | `Erc20` binding — a docstring and one `super().__init__(CHAIN_QNT)` |
| `allways/assets/__init__.py` | import, `__all__`, `ASSET_REGISTRY` row |
| `allways/constants.py` | `'qnt'` appended to `LAUNCH_SPOKES` |
| `.env.example`, `README.md` | Ethereum block now names three assets; `QNT_TOKEN_CONTRACT` |
| `tests/test_qnt_provider.py` | new, fully offline |
| `tests/test_chains.py` | registry case + the shared-network assertion covers both ETH tokens |
| `tests/test_cli_chain_networks.py` | fixture isolates `os.environ` (see below) |

[...]

PR #678, opened 2026-08-14
Disable spokes with no testnet deployment instead of failing neuron boot

anderdc:
## Problem

The testnet validator on `:test` crash-loops at startup:

```
RuntimeError: Qnt failed startup check: No qnt token contract for network 'sepolia' — set QNT_TOKEN_CONTRACT
```

QNT (#652) and PAXG (#670) are issuer-deployed on mainnet only — neither has a `TESTNET_TOKEN_CONTRACTS` entry (deliberately: QNT's test docstring notes a pinned test token must be immutable with no pause/blacklist, or the row's declared surface stops being true). But validators build every registry asset (`required_chains=None`), so the documented gap became a fleet-wide testnet boot crash. Fixing QNT alone just crashes on PAXG next.

## Fix

- `_token_contract` raises a typed `MissingTestnetDeployment` (a `ValueError` subclass) when a non-mainnet network has no pinned contract and no env override.
- `create_assets` catches it and **disables the spoke with a warning** instead of failing the boot — a spoke with no deployment on the network can't exist there at all, so failing the neuron over it protects nothing.
- A miner that **explicitly quotes the pair** (`chain_id in required_chains`) still fails hard — it must not advertise a pair it can't serve.
- `{ID}_TOKEN_CONTRACT` env override still works for pinning a mock; mainnet path (registry `asset_locator`) unchanged.

## Tests

- New: validator-style `create_assets(check=True)` boots and excludes the spoke; a quoting miner still raises.
- `test_qnt_provider.py` guard tests unchanged and passing (constructor still raises; the subclass keeps the `ValueError` contract); docstring updated to the new behavior.
- Full suite: 1857 passed; the one failure (`test_uppercase_bech32_matches_derived_address`) is a pre-existing ordering flake that reproduces on clean `test`.

## Deploy note

aw-vali does not auto-deploy (watchtower disabled) — after the `:test` image rebuilds, roll it manually. Until then the testnet vali can boot via the `QNT_TOKEN_CONTRACT`/`PAXG_TOKEN_CONTRACT` stand-ins already in its `.env`.

### DIFF
--- a/allways/assets/__init__.py
+++ b/allways/assets/__init__.py
@@ -11,7 +11,7 @@ from allways.assets.bnb import Bnb
 from allways.assets.btc import Bitcoin
 from allways.assets.chain import Chain
 from allways.assets.cro import Cro
-from allways.assets.erc20 import Erc20
+from allways.assets.erc20 import Erc20, MissingTestnetDeployment
 from allways.assets.eth import Ether
 from allways.assets.ethusdc import EthUsdc
 from allways.assets.evm_coin import EvmCoin
@@ -116,6 +116,13 @@ def create_assets(
             if check:
                 provider.check_connection(require_send=require_send)
             providers[chain_id] = provider
+        except MissingTestnetDeployment as e:
+            # A spoke with no deployment on this test network can't exist here at all, so it
+            # disables rather than failing the boot. A miner that quotes the pair (explicitly
+            # in required_chains) still fails hard — it must not advertise what it can't serve.
+            if check and required_chains is not None and chain_id in required_chains:
+                raise RuntimeError(f'{cls.__name__} failed startup check: {e}') from e
+            bt.logging.warning(f'{cls.__name__} disabled on this network: {e} — {chain_id} pairs are unavailable here')
         except Exception as e:
             if check and required:
                 raise RuntimeError(f'{cls.__name__} failed startup check: {e}') from e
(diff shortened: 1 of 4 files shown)


anderdc:
Update after checking for issuer testnet deployments:

**QNT now has a real testnet pin.** Quant runs an official test-QNT faucet on Sepolia — documented at docs.overledger.dev (`getTestQNT` on `0xCe8623CD5945dEcC23eC62D4A48aB90FF5D54825`, chainId 11155111). The token it dispenses, [`0x81Dc68CB065ec6D9a4d24f6e2F442dc2A236D853`](https://eth-sepolia.blockscout.com/address/0x81Dc68CB065ec6D9a4d24f6e2F442dc2A236D853), is a **verified plain ERC20** (QNT, 18 decimals): immutable, no pause/blacklist, and `paused()`/`isBlacklisted()`/`isFrozen()` all revert — the same surface the QNT row declares for mainnet. Added to `TESTNET_TOKEN_CONTRACTS` with provenance; qnt guard tests updated.

**PAXG stays disabled on testnet.** The only Sepolia "PAXG" ([`0x5A727f22…59224`](https://eth-sepolia.blockscout.com/address/0x5A727f226495E35d7F35D8d76977b33408c59224)) is an **unverified** third-party clone (creator `0xe0BbE1f5…`, name "PAX Gold Token" vs the real "PAX Gold") whose `paused()` reverts where real PAXG answers — pinning it would make testnet exercise a control surface mainnet doesn't have. It rides the new `MissingTestnetDeployment` degrade until we deploy a project-owned mock matching the declared surface.

PR #680, opened 2026-08-16
Red-team 08-14 fixes: fund-safety + slash-correctness pass

anderdc:
Verify-then-fix pass over the two 08-14 red-team audits (Claude + Codex), merged and worked one finding at a time. Every fix was traced in code before it was written, and each lands with tests. Threat model: honest, always-on validators ("code is law").

## Fixed (most → least severe)

**Critical**
- **C1 — cross-hub stale-claim reap** (`8514e84`): `close_stale_claim` was seeded on the reservation's own `collateral_chain` with no binding to the swap's backing, so a dual-hub miner's other-hub reservation could reap a healthy swap → permanent taker fund-strand, permissionless. Now `require!`s reservation backing == swap backing. (contract + LiteSVM test)
- **C2 — duplicate-source reservation theft** (`36f17a9`): a miner could back two live seats with one deposit across hubs. Reservation-phase, amount-agnostic `(from_chain, from_addr)` guard (deposit matching is `>=`, so the key must ignore amount).

**High**
- **H1 — dest == miner delivery addr** (`e8399d5`): reservation rejected when the user's `to_addr` equals the miner's delivery address.
- **H2 — SOL/TAO double-pay** (`5905be2`): own-broadcast dedup so a confirm timeout can't re-send a delivery.
- **H3+H4 — wrong-wallet delivery** (`dfaef89`): deliver from the swap's pinned `miner_to_addr` (not a last-write-wins cache) and refuse a send whose signing key ≠ pinned sender.
- **H6 — testnet neuron → mainnet spokes** (`538c2fa`): unset `{PREFIX}_NETWORK` fell through to the provider's mainnet default, so a testnet validator verified against mainnet / a miner could broadcast a real mainnet payout for a test obligation. Neurons now default unset spoke networks to the registry's `testnet_network` when `netuid != finney`; an explicit env var still wins.
- **H7 — slash on missing provider** (`d85daba`): a missing/disabled-spoke provider was treated as "tx absent" (slash-eligible) instead of "unverifiable". Now maps to SKIP, same as an unreachable RPC; the Active-overdue branch (which never fetches a leg) is guarded too.

**Medium / Low / rot**
- **M3 (partial, `538c2fa`)**: loud startup warning when a testnet neuron resolves a mainnet-looking Solana RPC. Hard program-id guard deferred.
- **M4 (`af8bf6f`)**: ERC-20 `check_connection` raises (not warns) on a token-contract probe failure.
- **L1 (`70216ba`)**: `polusdc` `min_onchain_amount` → 5 USDC, matching sibling USDC spokes.
- **L2 (`af8bf6f`)**: EVM gas-estimate revert detection via the structured `is_execution_revert` verdict, not a message substring.
- **R1 (`af8bf6f`)**: stale `arbusdc` unit-floor comments removed.

## Regression caught in-branch
`af8bf6f` (L2) correctly switched revert detection to the structured verdict but left six EVM delivery-gate test stubs raising a bare `RuntimeError`, which the new check doesn't recognize — the gates silently stopped bouncing reverting dests *in the tests* (production was always correct). Fixed in `d483fc4` by raising the real `EvmRpcError(code 3)` the node actually returns.

## Deliberately NOT in this PR
- **H5 (crown busy-forfeit leak)** — verified real; *not* a contract change (the chain already emits `reserved_until`), but the sound fix needs real crown-scoring machinery (append-only `activity_events` re-stamp + full replay re-test). Its own focused PR.
- **M1 (TAO bond-hotkey seizure)** — the one large on-chain fix; its own PR.
- **M2 / M5 / the L-tier / the hard half of M3** — Medium/Low follow-up batch.
- **ML3 (BTC `vin[0]` sender)** — pending a severity ruling (Codex High vs Claude Low).

## Tests
Full suite green: **1885 passed**. New coverage: cross-backing reap rejection (LiteSVM), duplicate-source refusal, dest==delivery rejection, SOL/TAO send dedup, pinned-sender delivery + wrong-wallet refusal, `apply_testnet_network_defaults`, and missing-provider-never-slashes across all swap statuses.

### DIFF
--- a/allways/__init__.py
+++ b/allways/__init__.py
@@ -1,3 +1,3 @@
-__version__ = '3.1.2'
+__version__ = '3.2.0'
 version_split = __version__.split('.')
 __spec_version__ = (1000 * int(version_split[0])) + (10 * int(version_split[1])) + (1 * int(version_split[2]))
--- a/allways/assets/btc.py
+++ b/allways/assets/btc.py
@@ -767,7 +767,7 @@ class Bitcoin(Asset, Chain):
         if override is not None:
             return max(1, override)
 
-        from allways.constants import BTC_FEE_RATE_SAFETY_MULTIPLIER, BTC_MIN_FEE_RATE
+        from allways.constants import BTC_FALLBACK_FEE_RATE, BTC_FEE_RATE_SAFETY_MULTIPLIER, BTC_MIN_FEE_RATE
 
         for attempt in range(2):
             try:
@@ -782,7 +782,10 @@ class Bitcoin(Asset, Chain):
                 bt.logging.debug(f'estimate_fee_rate: attempt {attempt + 1} failed: {e}')
             if attempt == 0:
                 time.sleep(3)
-        return BTC_MIN_FEE_RATE
+        # Estimation failed both attempts — broadcast at the survivable fallback, not the strand-prone
+        # floor (V-L5). Warn (not debug): a real send is going out at a non-estimated rate.
+        bt.logging.warning(f'estimate_fee_rate: /fee-estimates unavailable — using {BTC_FALLBACK_FEE_RATE} sat/vB fallback')
+        return BTC_FALLBACK_FEE_RATE
 
     def send_amount(
         self, to_address: str, amount: int, from_address: Optional[str] = None, dedup_key: Optional[str] = None
--- a/allways/assets/btc.py
+++ b/allways/assets/btc.py
@@ -784,7 +784,9 @@ class Bitcoin(Asset, Chain):
                 time.sleep(3)
         # Estimation failed both attempts — broadcast at the survivable fallback, not the strand-prone
         # floor (V-L5). Warn (not debug): a real send is going out at a non-estimated rate.
-        bt.logging.warning(f'estimate_fee_rate: /fee-estimates unavailable — using {BTC_FALLBACK_FEE_RATE} sat/vB fallback')
+        bt.logging.warning(
+            f'estimate_fee_rate: /fee-estimates unavailable — using {BTC_FALLBACK_FEE_RATE} sat/vB fallback'
+        )
         return BTC_FALLBACK_FEE_RATE
 
     def send_amount(
--- a/allways/assets/chain.py
+++ b/allways/assets/chain.py
@@ -55,8 +55,9 @@ class Chain(ABC):
         """Canonical comparison form for this chain's addresses (identity by default).
 
         Chains whose address encoding is case-insensitive (ETH hex / EIP-55 checksum casing)
-        override this so equality checks never fail on casing alone. Comparison only — never
-        feed the normalized form back on-chain or into display."""
+        override this so equality checks never fail on casing alone. SOURCE addresses commit
+        on-chain in this form (the source-lock PDA is byte-keyed on it — V-C2); delivery
+        addresses stay as given (EVM sends re-checksum at signing)."""
         return address
 
     @abstractmethod
--- a/allways/assets/erc20.py
+++ b/allways/assets/erc20.py
@@ -138,8 +138,11 @@ class Erc20(EvmAsset):
         try:
             code = self.chain.eth_rpc('eth_getCode', [self.token_contract, 'latest']) or '0x'
         except Exception as e:
-            bt.logging.warning(f'{self._log} token contract probe failed: {e}')
-            return
+            # For an ERC-20 the token contract IS the asset identity — a probe we can't complete
+            # must fail boot, not warn-and-continue (a warn boots a broken token as "ready").
+            raise ConnectionError(
+                f'{self.chain_def.id} token contract probe failed on {self.chain.network}: {e}'
+            ) from e
         if code == '0x':
             # A typo'd override or wrong network would otherwise surface as every send failing.
             raise ConnectionError(
--- a/allways/assets/erc20.py
+++ b/allways/assets/erc20.py
@@ -141,7 +141,9 @@ class Erc20(EvmAsset):
             # A probe we can't COMPLETE is transient (unreachable RPC / rate-limit storm) — degrade, never
             # crash the whole validator at boot over one spoke's flaky endpoint (that takes the hub down too).
             # The real config fault — a codeless/wrong contract — is a COMPLETED probe returning '0x' below.
-            bt.logging.warning(f'{self.chain_def.id} token contract probe unreachable on {self.chain.network} ({e}) — degraded')
+            bt.logging.warning(
+                f'{self.chain_def.id} token contract probe unreachable on {self.chain.network} ({e}) — degraded'
+            )
             return
         if code == '0x':
             # A typo'd override or wrong network would otherwise surface as every send failing.
--- a/allways/assets/erc20.py
+++ b/allways/assets/erc20.py
@@ -150,13 +150,20 @@ class Erc20(EvmAsset):
     def check_connection(self, require_send: bool = True) -> None:
         super().check_connection(require_send=require_send)
         try:
-            code = self.chain.eth_rpc('eth_getCode', [self.token_contract, 'latest']) or '0x'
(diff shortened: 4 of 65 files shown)


LandynDev:
Read every fix at the code. Here is what still needs to change.

**Fix before merge**

- `state.rs:421-423` — adding `hotkey` to `Swap` breaks every swap PDA that already exists. Six instructions stop decoding them, and the funds are stuck. Drain all live swaps before the upgrade, or ship a closer for the new layout.
- `extend_reservation.rs:42-47` — the new `SourceLock` account is required, so a reservation made before the upgrade can never extend. That is the slow-deposit case extend exists for. Use `init_if_needed`.
- `finalize_reservation.rs:99-102` — hash the normalized address, not the raw one. Today `0xABC…` and `0xabc…` make two different locks, and one deposit fills both. The CLI self-represented path has no check at all.
- `solana_swap_loop.py:432-438` — past `max_extend_at` a missing provider still slashes the miner. A spoke nobody wired is exactly the case that reaches quorum. Cancel no-fault instead of timing out.
- `tao.py:708` — 25 blocks is too short to call a transfer lost. A late inclusion gets paid twice. Widen it past the mortal era.
- `tao.py:736-743` — when the height read fails, the backfill re-anchors the scan on now and hides the real inclusion block. Same double pay. The 155 new TAO lines have no tests.

**Fixes that only went half way**

- `evm_coin.py:302` still falls back to 21000 gas while `:362` demands 100000, so a degraded estimator still costs the miner a slash. And `is_execution_revert` (`evm.py:30-32`) is still `'revert' in message`, so the comment at `:300` claims more than the code does.
- The testnet network default only reaches the neurons. The CLI never calls it and still defaults netuid to finney (`cli/main.py:170`).
- `solana/rpc.py:68` never blocks a mainnet neuron on a testnet cluster. Also `apply_testnet_network_defaults` sets `BTC_NETWORK=testnet4`, which moves local dev off regtest.
- `PoolOpened` and `ReservationRequested` are still unmapped, so the crown index still over-credits. And `event_index.py:126-133` inserts an expiry instead of moving it, which frees the next reservation early.
- `btc.py:332-345` still never fails over to mempool.space on a 200. The fallback is a guess, not an estimate.
- `erc20.py:481-489` does not probe the fee, so users still enter swaps that are guaranteed to cancel.
- `tao.py:721` and `btc.py:670` compare raw address strings. Use `normalize_address` — it already exists and the EVM scanner uses it.

**Smaller**

- Clear the source lock in `vote_initiate` and `close_stale_claim`. Right now only time clears it, so a repeat swap from the same wallet onto the same miner fails for the rest of the TTL and burns the reservation fee.
- `SourceLock` is never closed. That is permanent rent per address, paid by the router.
- `bind_hotkey.rs:97-100` — a typo now burns the pubkey as well as the hotkey, and recovery resets `successful_swaps`. Add an escape hatch.
- `state_store.py:239-249` — add a lower bound on the row it moves, or it can clobber an old expiry and strand a hub.
- `fulfillment.py:170` still checks the old wallet while `:288` spends the pinned one.
- `erc20.py:153` — `or '0x'` turns a null RPC answer into a boot failure. That is the case `e6eb00a` meant to prevent.
- `vote_initiate.rs:155` — `unwrap_or_default()` hides a decode failure as a zero hotkey.
- `solana_swap_loop.py:433` — `max_extend_at == 0` times out on the first pass.
- `relay/slash.py:33` — prefer the on-chain pin over the local snapshot. The pin is the value every validator shares.
- `sol.py:381` and `tao.py:763` evict unresolved sends, and `tao.py:733` overwrites them. Both remove the guard against a re-send.
- Upgrade the validator fleet together. Three instructions changed their accounts or args, so a mixed fleet stalls quorum.

**Tidy**

- `SRCLOCK_SEED` has no `#[constant]`, so it misses the IDL and clients hardcode `b"srclock"`.
- `state.rs:357` and `:362` say opposite things about when the lock is free.
- `state.rs:378` says `vote_initiate` creates the Swap. `submit_swap_claim.rs:44` does.
- `neurons/miner.py:45-54` and `neurons/validator.py:80-89` are identical, and they guard in a different order.
- `tao.py` has three block-walk loops with three different windows. That divergence is where the double pay lives.
- `event_index.py:70-75` still says unbound-pubkey events are dropped. `92340d7` changed that.

Everything else in the PR held up.

anderdc:
Thanks for the thorough pass — every point got verified against the tree and dispositioned. Since your review, 7 commits landed on the branch (plus the 3.2.0 bump).

**Fixed on-branch**
- `tao.py:708` — you were right, and it was the best catch in the review: the 25-block dead threshold sat inside bittensor's 128-block mortal era. Now `LOST_SEND_DEAD_BLOCKS = 130`, used only for the dead declaration. (This also resolves the "three windows" point — the finality deadline has its own constant now.)
- Source-lock normalization — fixed, but not by hashing the normalized form (the program re-hashes `user_from_addr` on-chain, so that alone would break finalize). The *address itself* is canonicalized at both intakes + the sweep commit, and `confirm_deposit` now picks deterministically among multi-matched slots: canonical-form wins, then oldest. A variant can only come from bypassing our tooling, and byte-exact dupes are lock-blocked.
- `erc20.py:153` — confirmed, null getCode now degrades instead of crashing boot.
- `event_index.py:126-133` — the reap now *moves* the draw+ttl guess instead of inserting beside it, and `restamp_reservation_expiry` gained a `not_before` floor so an orphaned Filled/Extended can't drag a prior reservation's fired row (your `state_store.py:239` point).
- Source lock now **releases at initiate quorum** (`vote_initiate` +`from_addr_hash` +lock account) — kills the same-wallet repeat-swap fee burn. `close_stale_claim` stays clock-freed (its lock is nearly expired by definition).
- All four tidy items (SRCLOCK `#[constant]`, both state.rs comments, the ingest docstring).

**Queued as follow-ups (agreed, not merge-blocking)**
Reserve-time fail-closed on a missing provider (better than cancel-at-ceiling — cancel would hand the miner the deposit); getCode-gated defer for the 21000 fallback (blanket defer strands EOA payouts in an outage); BTC fee failover on a degenerate 200; entry-time `getFeeFor` probe; the fulfillment gate reading the pinned wallet; deduping the neuron guard blocks.

**Declined, with the facts that decided it**
- Swap PDA decode: `hotkey` is appended *last* and old accounts carry >32 bytes of string padding, so they decode with `hotkey = [0;32]` — exactly the pre-upgrade sentinel `slash.py` handles. The migrate driver also hard-refuses while any Swap PDA exists. (It's 7 instructions that read Swap, not 6.)
- Extend-on-pre-upgrade-reservations: real, but covered by deploy discipline — we halt and wait out the reservation TTL before upgrading; added to the upgrade runbook.
- "155 new TAO lines have no tests": 18 test functions landed with 59e02b1 covering exactly those failure paths (unreadable head/block raise, anchor-on-recorded-block, backfill, lost-hash resolution).
- TAO backfill re-anchor: a failed head read at backfill raises-and-waits — it never re-sends. The surviving residual is documented and shrinks further with the 130-block widening.
- `normalize_address` on TAO is the identity function — the suggested call is a no-op; btc.py:670's both sides come from the miner's own tooling.
- `is_execution_revert` is `code == 3 OR substring` — the structured check is load-bearing; softened the over-claiming comment.
- BTC_NETWORK: the prior unset default was mainnet, not regtest (regtest isn't in CHAIN_BTC.networks).
- `max_extend_at == 0`: unreachable — only Active/Fulfilled rows hit that branch and vote_initiate sets it nonzero in the same instruction.
- Set-once binding escape hatch / SourceLock rent / pin-vs-snapshot order: deliberate decisions, unchanged (the freeze is what makes the seizure un-dodgeable; under it, snapshot and pin can only disagree when one is empty).

**Deploy:** mixed-fleet point taken — the footprint is now four changed surfaces (finalize/extend/vote_initiate + Swap layout); full runbook written (halt → wait out reservation TTL → drain swaps → coordinated restart). Version bumped to 3.2.0.

Ready for another look.

PR #700, opened 2026-08-21
feat(btc): OAuth client-credentials auth for Blockstream Explorer API

anderdc:
## Why
Maestro emailed that its Bitcoin API shuts down on **2026-09-18**. It is the paid first rung in prod's `BTC_ESPLORA_URLS`. Blockstream Explorer API (native Esplora, Advanced 1M credits/\$100) replaces it; Dwellir and SN19 Blockmachine have no Bitcoin, mempool.space enterprise starts at \$1,499/mo, and every other hosted BTC vendor sells Core JSON-RPC/Blockbook, not the Esplora dialect `btc.py` speaks.

Blockstream's keyed endpoint authenticates with OAuth2 `client_credentials`: tokens expire after 300 s and there is no refresh token, so our static `url|key` → `Authorization: Bearer` scheme can't be used unmodified.

## What
- New `BTC_ESPLORA_URLS` entry form `url|oauth:CLIENT_ID:CLIENT_SECRET` → `OAuthClientCredentials`, which mints a bearer lazily, caches it, and refreshes 60 s before expiry.
- `btc_api_request`: an OAuth rung that returns 401/403 invalidates the token and retries **that rung once** before the existing failover; a token-endpoint failure is treated like any request error (fails over to the next rung).
- Static `url|key` entries, `BTC_ESPLORA_API_KEY_HEADER`, and the public blockstream → mempool defaults are unchanged. No response parsing / verify-path code touched.
- `.env.example` documents the three entry forms.

## Tests
`tests/test_btc_esplora_oauth.py` (9): parsing (static unchanged, oauth, secret containing `:`, malformed), token minted once and reused, refresh inside the margin, stale-token 401 → refresh + same-rung retry, persistent 401 → one retry then failover, token endpoint 503 → failover, static-key 401 does not retry. Full BTC suite: 111 passed. Live probe of Blockstream's token endpoint with bogus creds returns `invalid_client` (request shape accepted).

## Rollout
1. Blockstream dashboard → Manage Keys → one key per environment.
2. lena: `BTC_ESPLORA_URLS=https://enterprise.blockstream.info/api|oauth:ID:SECRET,https://mempool.space/api,https://blockstream.info/api`; restart; watch `[Esplora]` logs.
3. isaiah (testnet4): Blockstream documents only mainnet + testnet3 — keep public mempool testnet4 unless the dashboard offers testnet4.
4. Remove the Maestro entry before 9/18 either way.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,10 +30,12 @@ SOLANA_RPC_URL=
 
 # ─── Chain: Bitcoin (BTC pairs) ────────────────────────────
 BTC_NETWORK=mainnet                 # mainnet | testnet | testnet4
-# OPTIONAL ordered esplora endpoints, URL or URL|API_KEY; unset = public (blockstream, mempool.space)
-# keyed entries send the key via the header below; "Authorization" adds a Bearer prefix
-# BTC_ESPLORA_API_KEY_HEADER=api-key
-# BTC_ESPLORA_URLS=https://host/esplora|your_key,https://mempool.space/api
+# OPTIONAL ordered esplora endpoints; unset = public (blockstream, mempool.space). Entry forms:
+#   URL                              anonymous
+#   URL|API_KEY                      static key via BTC_ESPLORA_API_KEY_HEADER ("Authorization" adds Bearer)
+#   URL|oauth:CLIENT_ID:CLIENT_SECRET Blockstream Explorer API (5-min tokens, auto-refreshed)
+# BTC_ESPLORA_API_KEY_HEADER=Authorization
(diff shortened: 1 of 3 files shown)


anderdc:
Live-tested with a real Blockstream key (mainnet + testnet3 tip/fee-estimates via the ladder, stale-token 401 → refresh → same-rung retry all 200). Found that Blockstream answers **402 Payment Required** for networks the plan lacks (testnet4) — and presumably when credits run out — and `failover_reason` returned that to the caller instead of falling through. Added 402 to the failover set + test; testnet4 ladder now falls back to public mempool live.

PR #708, opened 2026-08-25
Add SN7 and SN74 subnet alpha assets with hub pairs (PR2)

LandynDev:
Add SN7 and SN74 subnet alpha as swappable assets, paired with the hubs only (`sol<->sn7`, `tao<->sn7`, `sol<->sn74`, `tao<->sn74`). Stacked on #707.

Registry
- `CHAIN_SN7` / `CHAIN_SN74`: TAO's env prefix, clock and reorg depth; `netuid` + `backing_family='tao'`. `constants.family()` is the Python twin of `backing.rs::family` (prefix rule; a test pins registry rows to it). `LAUNCH_ALPHAS` launches the hub pairs.
- `hub_leg` / `declarable_backings` / `canonical_pair` learn the family rule: the literal hub anchors, else the family-bearing leg (an alpha is its own scoring family, spec §6). `LAUNCH_PAIRS` stays in canonical order (pinned). Alpha<->spoke pairs are deliberately absent — they mint new scoring families and are gated behind #709.

Asset
- `assets/alpha.py`: first real use of the Asset/Chain seam — `Alpha` binds the existing `Tao` as its chain. Settlement = a top-level `SubtensorModule.transfer_stake` (matched by name, no pallet indices) whose extrinsic emitted `ExtrinsicSuccess` AND `StakeTransferred`, with the alpha amount read from the call args (the event carries the TAO-equivalent). Sends pick the hotkey holding the most alpha, `mev_protection=False`, per-asset dedup.
- Tao's block-window scan, cursor scanner and prior-send resolution are lifted into chain-level methods parametrized by the asset's decoder; Tao's own paths are thin, behaviour-identical wrappers.
- `can_deliver_to` / `delivery_refused` / `cancel_evidence` read `TransferToggle` and `SubtokenEnabled` (subnet-level, not user-weaponizable — spec §4.5).

Cover check (spec §5)
- `swap_intake.leg_value` is the Python twin of `LegBind`: exact legs first, a declared alpha leg is valued at spot via `Alpha.value_rao`. Same function sizes `collateral_amount` at intake, gates the user's deposit, and gates the validator's initiate vote. Binary, no tolerance.

CLI: `alw miner quotes --sn7-price` / `--sn74-price` derive from `LAUNCH_ALPHAS`; the alpha address IS the TAO address — no new flags, no new env vars.

Tests: `uv run pytest` (full suite) + `ruff`. Testnet validates the full mechanism on `sn7<->tao` (spec §9).

Emissions: hub pairs land inside the existing sol/tao families — 31 -> 35 pairs, alpha share 0%.

### DIFF
--- a/allways/assets/__init__.py
+++ b/allways/assets/__init__.py
@@ -2,6 +2,7 @@ from typing import Dict, NamedTuple, Optional, Set, Tuple, Type
 
 import bittensor as bt
 
+from allways.assets.alpha import Alpha
 from allways.assets.arbusdc import ArbUsdc
 from allways.assets.asset import Asset, MissingTestnetDeployment, SendResult, TransactionInfo
 from allways.assets.aster import Aster
@@ -20,6 +21,8 @@ from allways.assets.paxg import Paxg
 from allways.assets.pol import Pol
 from allways.assets.polusdc import PolUsdc
 from allways.assets.qnt import Qnt
+from allways.assets.sn7 import Sn7
+from allways.assets.sn74 import Sn74
 from allways.assets.sol import Sol, SolanaChain
 from allways.assets.solusdc import SolUsdc
 from allways.assets.spl_token import SplToken
@@ -54,6 +57,9 @@ __all__ = [
     'SolanaChain',
     'SplToken',
     'SolUsdc',
+    'Alpha',
+    'Sn7',
+    'Sn74',
     'create_assets',
 ]
 
@@ -86,6 +92,8 @@ ASSET_REGISTRY: Tuple[AssetSpec, ...] = (
     AssetSpec('polusdc', PolUsdc, ()),
     AssetSpec('paxg', Paxg, ()),
     AssetSpec('solusdc', SolUsdc, ('solana_rpc_url', 'solana_keypair')),
+    AssetSpec('sn7', Sn7, ('subtensor', 'wallet')),
+    AssetSpec('sn74', Sn74, ('subtensor', 'wallet')),
 )
 
 
--- a/allways/assets/alpha.py
+++ b/allways/assets/alpha.py
@@ -231,8 +231,10 @@ class Alpha(Asset):
         tx_hash = getattr(receipt, 'extrinsic_hash', None) or Tao.extrinsic_hash(getattr(response, 'extrinsic', None))
         if tx_hash:
             self.broadcasted_txids[scope] = (to_address, int(amount), tx_hash, attempt_head)
-        if not response.success:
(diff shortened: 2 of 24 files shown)


LandynDev:
Pushed a2970d9 — decoder correctness, found reviewing this against the #606 settlement lesson.

- `origin_netuid` is now pinned alongside `destination_netuid`. `alpha_amount` is denominated in the origin subnet, and only a within-subnet transfer moves alpha 1:1 — a cross-netuid call unstakes through the AMM and lands a different amount of ours, so crediting it paid out on funds that never arrived.
- A raw block now raises instead of decoding as absent. The raw fallback parses Balances transfers only, so every stake transfer read as "no such payment" — a slash-eligible verdict on a dest leg, and a re-send on a send-landed check.
- `transfer_stake_and_hotkey` is accepted. It changes the owning coldkey through the same dispatch path, so it settles 1:1 within a subnet too; only its settlement event differs. Hotkeys stay unread — they change who takes a delegate cut, never ownership or the amount.

Three regression tests. The existing `_ext()` helper set both netuids from one argument, so nothing could have caught the cross-netuid case; it now takes them separately.

Still open on this stack: this branch is 24 commits behind `test` (8 conflicting files), and `CANCEL_REASON_ALPHA_TRANSFER_DISABLED = 6` on #707 collides with `CANCEL_REASON_INVALID_DEST = 6` that landed on `test` — needs renumbering in the Rust and the Python mirror.

LandynDev:
Pushed 5d2b7d6 — all 128 subnet alphas now register from one generated row.

An alpha is the same asset with a different netuid, so a hand-written `ChainDefinition` per subnet was 128 copies of one paragraph, and the per-netuid subclasses existed only to bind that row. The rows generate from `ALPHA_NETUIDS`, and each binds with a `partial` — which deletes `sn7.py` and `sn74.py` and makes the 129th subnet a range change.

**Registered is not launched.** All 128 resolve as assets; `LAUNCH_ALPHAS` still decides which get pairs and emissions, and it stays at sn7/sn74 until #740. Shipping 128 into `LAUNCH_PAIRS` would dilute every existing pair's zero-volume floor about ninefold, which is exactly what #740 exists to answer.

`transfers_enabled` now also reads `NetworksAdded`. `SubnetLimit` is full on finney (128 subnets, no gaps), so the next registration dissolves the lowest-priced subnet and force-liquidates its alpha to coldkey TAO. A miner then cannot deliver, and that is not its fault — it joins `SubtokenEnabled` and `TransferToggle` as chain-refused rather than a slash.

2593 tests pass, ruff clean.

PR #723, opened 2026-09-05
Hard-exit the neuron when its forward loop stalls

LandynDev:
The stall watchdog in neurons/validator.py and neurons/miner.py calls sys.exit(1), but a stalled forward thread is wedged inside a socket read that never returns. The interpreter then waits for that thread at shutdown (concurrent.futures atexit join), so the process never exits, docker's restart policy never fires, and the neuron sits dead with the container reported Up.

Seen three times on the testnet validator today: the relay's vault event poll blocks forever in websockets recv under async_substrate_interface (stack in the linked issue), the watchdog logs "exiting for restart", the thread stops, the process stays alive.

exit_for_restart() logs, arms a daemon Timer that calls os._exit(1) after 30 s, then raises SystemExit so the normal cleanup still runs when it can. Both neurons use it for the dead-thread and stalled-forward cases.

Stopgap only. The root cause (no read timeout on the vault subtensor websocket in allways/vault/client.py poll_events) is filed separately.

### DIFF
--- a/neurons/base/neuron.py
+++ b/neurons/base/neuron.py
@@ -1,5 +1,7 @@
 import copy
 import os
+import sys
+import threading
 import time
 from abc import ABC, abstractmethod
 from typing import Callable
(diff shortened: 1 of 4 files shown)


LandynDev:
Closing: the fix belongs on the RPC connection (timeout + reconnect on the vault poll, #724), not a hard-exit workaround.

